PR #6, opened 2022-02-17
[POL-182, POL-170] Set up github workflows

snwessel:
Copying the tag and publish workflows from ECM.


dvoet (approved):
these look fine but can we pull in the build and test workflow too?

snwessel:
Note: I added in the build and test workflow, not including the "integration" and "sonarcloud" steps since they'll obviously require a bit more setup.

Shakespeared (approved):

PR #9, opened 2022-03-03
simplify & dedupe 

zarsky-broad:
- Removing checks for conditions that can never happen because they've already been checked
- Using Apache commons' `isEmpty`, it's null-guarded
- move `getFenceAccessToken` inside `getAccessUrl`


Shakespeared (approved):
LGTM

dvoet (approved):

PR #16, opened 2022-03-22
add testrunner setup [POL-187]

Shakespeared:


dvoet (approved):

integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -0,0 +1,60 @@
...
+      throw new IllegalArgumentException("Service URI cannot be empty");
+    }
+
+    var apiClient = new ApiClient();
+    apiClient.setBasePath(server.policyManagerUri);
+    return apiClient;
+  }
+
+  /**
+   * Build the ECM API client object for the given test user and server specifications. The test
```
snwessel:
```suggestion
   * Build the DRS Hub API client object for the given test user and server specifications. The test
```

integration/src/main/java/scripts/utils/ClientTestUtils.java
```diff
@@ -0,0 +1,60 @@
...
+public class ClientTestUtils {
+
+  private ClientTestUtils() {}
+
+  // Required scopes for client tests include the usual login scopes and GCP scope.
+  public static final List<String> TEST_USER_SCOPES =
+      List.of("openid", "email", "profile", "https://www.googleapis.com/auth/cloud-platform");
+
+  /**
+   * Build the ECM API client object for the server specifications. No access token is needed for
```
snwessel:
```suggestion
   * Build the DRS Hub API client object for the server specifications. No access token is needed for
```

integration/src/main/resources/servers/drshub-local.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "drshub-local",
+  "description": "drshub Local env.",
+
+  "samUri": "https://sam.dsde-dev.broadinstitute.org",
+  "policyManagerUri": "http://localhost:8080",
```
snwessel, line 6: Actually, this `policyManagerUri` value is supposed to be specific to ECM - and each service is supposed to have its own URI field in the TestRunner configs. It's a bit silly in my opinion, and there's been a lot of back and forth in the #testrunner slack channel about changing that approach. For now we probably need to make a quick change in the the test runner codebase (similar to [this](https://github.com/DataBiosphere/terra-test-runner/pull/29/files)) to give drshub its own URI config value.

snwessel (approved):
I just realized that there needs to be one quick change in the testrunner code base before we call this complete. Otherwise, this looks great.

integration/src/main/resources/servers/drshub-local.json
snwessel: If we make a PR for that on the testrunner repo, it may also be worthwhile to rename the ECM config to a more clear name than "policyManagerURI".

PR #17, opened 2022-03-29
[POL-181] Add missing tests

snwessel:
Wrapping up testing for the DrsHub:
- Adding unit tests for `ShouldFetchFenceAccessToken`, since I could tell from the code coverage that it was only tested in the scenario where it returns "true".
- Removing number comments from the controller tests
- Adding tests for the getVersion and getStatus endpoints (these tests are just copied from ECM).

Let me know if there are any other unit tests you all think we should be adding!


dvoet (approved):

service/src/test/java/bio/terra/drshub/config/DrsProviderTest.java
```diff
@@ -0,0 +1,82 @@
...
+      ProviderAccessMethodConfig.create()
+          .setType(AccessMethodConfigTypeEnum.gs)
+          .setAuth(AccessUrlAuthEnum.fence_token)
+          .setFetchAccessUrl(true);
+
+  @Test
+  void testShouldFetchFenceAccessTokenReturnsTrue() {
+    // returns true when there are no problems identifying the relevant configs
+    var accessMethods = new ArrayList<ProviderAccessMethodConfig>();
+    accessMethods.add(testAccessMethodConfig);
```
zarsky-broad: Can you use `List.of` for these?

service/src/test/java/bio/terra/drshub/config/DrsProviderTest.java
```diff
@@ -0,0 +1,82 @@
...
+    DrsProvider drsProvider = createRandomDrsProvider().setAccessMethodConfigs(accessMethods);
```
zarsky-broad: `var`?

service/src/test/java/bio/terra/drshub/config/DrsProviderTest.java
snwessel: I agree that would be tidier. I did try that and the issue is that `List.of` method outputs a `List` but our [DrsProviderInterface config field](https://github.com/DataBiosphere/terra-drs-hub/blob/1f454b6afb1dd4bd311d4a8151083deed3886b5d/service/src/main/java/bio/terra/drshub/config/DrsProviderInterface.java#L27) expects an `ArrayList`. If you change the `DrsProviderInterface` to expect a List instead, it complains that it fails to "load the ApplicationContext" because it "cannot bind the property" because there's more than one implementation for the List interface. 

If you have any thoughts on how to get around that, I'd be happy to look into them. Otherwise, I might just leave this as is.

service/src/test/java/bio/terra/drshub/config/DrsProviderTest.java
snwessel: Good point, thank you! Just fixed.

service/src/test/java/bio/terra/drshub/config/DrsProviderTest.java
zarsky-broad: I found this works:
```java
var accessMethods = new ArrayList<>(List.of(testAccessMethodConfig));
```

Also, could we declare `accessMethods` once as `final`?

snwessel: Yeah, that sounds like a good plan! Thanks for the suggestion!

snwessel: Let me know if the changes I made are what you had in mind.

zarsky-broad (approved):

PR #21, opened 2022-04-01
POL-199 Integration tests

dvoet:
I think this is the only integration test we need to write. All the others test error conditions or mock DRS servers, both of which should be covered by unit tests. The real integration test is manual https://docs.google.com/document/d/1-SXw-tgt1tb3FEuNCGHWIZJ304POmfz5ragpphlq2Ng/edit#.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -3,11 +3,11 @@ info:
   title: drsHub
   version: 2.0.0
 paths:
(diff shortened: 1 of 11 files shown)


snwessel (approved):
LGTM! And I agree, `resolveDrs`does seem like a much better name!

PR #34, opened 2022-07-07
ID-84 Sentry config and stackdriver logging support

tlangs:
https://broadworkbench.atlassian.net/browse/ID-84

Adding Sentry and Stackdriver logging support. The `SENTRY_DSN` environment variable still needs to be populated in deployed environments, but that will happen through terra-helmfile. 

This also adds support for human-readable logging. When running locally, set `DRSHUB_LOG_APPENDER=Console-Standard` in your env, and DrsHub will log normal lines instead of Stackdriver JSON. Useful for local development!

### DIFF
--- a/service/src/main/resources/logback.xml
+++ b/service/src/main/resources/logback.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--We need this file to override the terra-commons-lib configuration for human-readable logging-->
+<configuration scan="true" scanPeriod="30 seconds">
+
+    <variable name="LOG_PATH" value="logs"/>
+
+    <if condition='!isDefined("DRSHUB_LOG_APPENDER")'>
+        <then>
+            <variable name="DRSHUB_LOG_APPENDER" value="Console-Stackdriver" />
+        </then>
+    </if>
+
+    <appender name="Console-Standard" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
+            <Pattern>%date %-5level [%thread] %logger{36}: %message%n</Pattern>
+        </encoder>
+    </appender>
+
+    <appender name="Console-Stackdriver" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
+            <layout class="org.springframework.cloud.gcp.logging.StackdriverJsonLayout">
+                <includeTraceId>true</includeTraceId>
+                <includeSpanId>true</includeSpanId>
+            </layout>
+        </encoder>
+    </appender>
+
+    <if condition='isDefined("SENTRY_DSN")'>
+        <then>
+            <appender name="Sentry" class="io.sentry.logback.SentryAppender"/>
+        </then>
+    </if>
+
+    <root level="INFO">
+        <appender-ref ref="${DRSHUB_LOG_APPENDER}"/>
+        <if condition='isDefined("SENTRY_DSN")'>
+            <then>
+                <appender-ref ref="Sentry"/>
+            </then>
+        </if>
+    </root>
+</configuration>
--- a/README.md
+++ b/README.md
@@ -1,0 +1,6 @@
+# Dr. Shub, MD
+## Also known as DrsHub
+
+## Logging
+By default, DrsHub will emit logs in the Stackdriver JSON format. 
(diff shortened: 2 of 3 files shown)


Shakespeared (approved):

service/src/main/resources/logback.xml
```diff
@@ -0,0 +1,42 @@
...
+                <includeSpanId>true</includeSpanId>
+            </layout>
+        </encoder>
+    </appender>
+
+    <if condition='isDefined("SENTRY_DSN")'>
+        <then>
+            <appender name="Sentry" class="io.sentry.logback.SentryAppender"/>
+        </then>
+    </if>
```
andy7i, line 32: > Adding Sentry and Stackdriver logging support. The SENTRY_DSN environment variable still needs to be populated in deployed environments, but that will happen through terra-helmfile.

found the corresponding PR:https://github.com/broadinstitute/terra-helmfile/pull/3033/files

andy7i: I wasn't sure what the Sentry DSN was so I looked it up and found an answer:

> Data Source Name (DSN)
> Sentry automatically assigns you a Data Source Name (DSN) when you create a project to start monitoring events in your app.
> 
> What the DSN Does
> A DSN tells a Sentry SDK where to send events so the events are associated with the correct project.
> 
> If this value is not provided, SDKs will try to read it from the SENTRY_DSN environment variable, where applicable. This fallback does not apply in cases like a web browser, where the concept of environment variables does not exist.
> 
> If an SDK is not initialized or if it is initialized with an empty DSN, the SDK will not send any data over the network, such as captured exceptions.
> 
>

service/src/main/resources/logback.xml
```diff
@@ -0,0 +1,42 @@
...
+<!--We need this file to override the terra-commons-lib configuration for human-readable logging-->
+<configuration scan="true" scanPeriod="30 seconds">
+
+    <variable name="LOG_PATH" value="logs"/>
+
+    <if condition='!isDefined("DRSHUB_LOG_APPENDER")'>
+        <then>
+            <variable name="DRSHUB_LOG_APPENDER" value="Console-Stackdriver" />
+        </then>
+    </if>
```
andy7i, line 11: > This also adds support for human-readable logging. When running locally, set DRSHUB_LOG_APPENDER=Console-Standard in your env, and DrsHub will log normal lines instead of Stackdriver JSON. Useful for local development!

That's awesome! Could we add this to the README?

service/src/main/resources/logback.xml
```diff
@@ -0,0 +1,42 @@
...
+        </encoder>
+    </appender>
+
+    <if condition='isDefined("SENTRY_DSN")'>
+        <then>
+            <appender name="Sentry" class="io.sentry.logback.SentryAppender"/>
+        </then>
+    </if>
+
+    <root level="INFO">
```
andy7i, line 34: Looking up about this root element, I found: 
> The possible levels are, in order of precedence: TRACE, DEBUG, INFO, WARN and ERROR. Each level has a corresponding method that we use to log a message at that level. If a Logger isn't explicitly assigned a level, it inherits the level of its closest ancestor. The root logger defaults to DEBUG.

So this means we are logging INFO, WARN, and ERROR, but not DEBUG or TRACE right? Would we want DEBUG for local logs?

service/src/main/resources/logback.xml
```diff
@@ -0,0 +1,42 @@
...
+            <layout class="org.springframework.cloud.gcp.logging.StackdriverJsonLayout">
+                <includeTraceId>true</includeTraceId>
+                <includeSpanId>true</includeSpanId>
+            </layout>
+        </encoder>
+    </appender>
+
+    <if condition='isDefined("SENTRY_DSN")'>
+        <then>
+            <appender name="Sentry" class="io.sentry.logback.SentryAppender"/>
```
andy7i, line 30: Maybe something we could/should verify is that the output pattern/layout formatting is reasonable?

andy7i (approved):
This is new to me so I had a few questions. Overall looks very reasonable and nothing is blocking!

service/src/main/resources/logback.xml
tlangs: We might want DEBUG for local logs, but you can always just edit the xml file to do so. I'm hesitant to emit DEBUG logs by default because that can result in a TOOOON of logs that we're not always interested in.

service/src/main/resources/logback.xml
tlangs: What README??? 🤣 I'll create a README  and add a logging section

PR #39, opened 2022-07-19
ID-156 Don't log status and version endpoint calls

tlangs:
K8s engines calls the status endpoint of apps several times a second, and the version endpoint once per minute. These logs are pretty useless to us, and just contribute to the amount of log messages needing to be ingested by logit and parsed through by stackdriver. It decreases our ability to diagnose problems (although we could just ignore certain messages, but we’d still be paying for them). By excluding the status and version calls from the k8s engine, we can save ~260,000 log messages a day! In each environment! For each app!!!

I've retained the structure of the existing request logging, but augmented it to warn on 4xx errors, and include a bunch of extra info in the stackdriver json payload.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/config/WebConfig.java
+++ b/service/src/main/java/bio/terra/drshub/config/WebConfig.java
@@ -1,0 +1,17 @@
+package bio.terra.drshub.config;
+
+import bio.terra.drshub.logging.LoggerInterceptor;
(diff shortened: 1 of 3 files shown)


Shakespeared (approved):

andy7i (approved):
forgot to mention in the other PR, but I'm excited to be able to use all this extra information when we need to debug an issue!

PR #41, opened 2022-07-27
ID-17 Add Audit Logging and Log DRS Resolutions

tlangs:
Adding an AuditLogger and logging all successful and failed DRS resolutions in one place.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/MetadataService.java
+++ b/service/src/main/java/bio/terra/drshub/services/MetadataService.java
@@ -146,11 +146,24 @@ public record MetadataService(
       BearerToken bearerToken,
       boolean forceAccessUrl,
       String ip) {
+
+    AuditLogEvent.Builder auditEventBuilder =
+        new AuditLogEvent.Builder()
+            .dRSUrl(uriComponents.toUriString())
+            .providerName(drsProvider.getName())
+            .clientIP(Optional.ofNullable(ip));
     var drsMetadataBuilder = new DrsMetadata.Builder();
 
-    var drsResponse =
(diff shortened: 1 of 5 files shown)


Shakespeared (approved):

service/src/main/java/bio/terra/drshub/services/MetadataService.java
```diff
@@ -139,12 +144,26 @@ private DrsMetadata fetchMetadata(
...
       BearerToken bearerToken,
-      boolean forceAccessUrl) {
+      boolean forceAccessUrl,
+      String ip) {
+
+    AuditLogEvent.Builder auditEventBuilder =
+        new AuditLogEvent.Builder()
+            .dRSUrl(uriComponents.toUriString())
+            .providerName(drsProvider.getName())
+            .clientIP(Optional.ofNullable(ip));
```
andy7i, line 154: IP addresses aren't necessarily the most accurate identifier so I have some concerns about our ability to audit these requests. Is this our only option?

andy7i (approved):
love this clean, simple approach! One comment/concern but assuming we can't do anything about it, here's a thumb: 👍

service/src/main/java/bio/terra/drshub/services/MetadataService.java
tlangs: This is our only option. I know its not the most accurate identifier, but we don't currently get the `SamUser` for the request, and doing so for every request would add a loooot of calls over the network to Sam, especially since DRSHub should be handling tons of requests for analysis workflows. There could be hundreds of thousands of DRS requests at the start of a large workflow! 

We could probably add a cache in front of the user getting, but that's more complexity. If IP is good enough for ECM, its probably good enough here too!

PR #43, opened 2022-07-29
[ID-122] enable signed urls for BDC and CRDC

Shakespeared:
[[ID-122](https://broadworkbench.atlassian.net/browse/ID-122)]

one line change to enable signed urls for BDC and CRDC and make sure drshub config is the same as current martha config state

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -32,7 +32,7 @@ drshub:
       accessMethodConfigs:
         - type: gs
           auth: fence_token
-          fetchAccessUrl: false
+          fetchAccessUrl: true
     crdc:
       name: NCI Cancer Research / Proteomics Data Commons (CRDC / PDC)
       hostRegex: '.*\.datacommons\.io'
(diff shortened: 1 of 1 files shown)


tlangs (approved):

service/src/main/resources/application.yml
```diff
@@ -32,7 +32,7 @@ drshub:
       accessMethodConfigs:
         - type: gs
           auth: fence_token
-          fetchAccessUrl: false
+          fetchAccessUrl: true
     crdc:
```
tlangs, line 36: Wasn't this ticket to do CRDC too?

tlangs (changes requested):

service/src/main/resources/application.yml
Shakespeared: Yeah sorry I was looking into this at the very end of the day yesterday and got into the weeds with trying to make sure these configs match and then forgot to actually say what I was doing. But long story short, I was right the first time and yes this should be updated.

andy7i (approved):

tlangs (approved):

PR #44, opened 2022-08-10
ID-112 Method to fetch DRS authorizations + test

tlangs:
Simple as that. What we do with the response will be finished in later tickets.

https://broadworkbench.atlassian.net/browse/ID-112

### DIFF
--- a/service/src/test/java/bio/terra/drshub/BaseTest.java
+++ b/service/src/test/java/bio/terra/drshub/BaseTest.java
@@ -1,14 +1,37 @@
 package bio.terra.drshub;
 
+import static bio.terra.drshub.controllers.DrsHubApiControllerTest.TDR_TEST_HOST;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import bio.terra.drshub.config.DrsHubConfig;
+import bio.terra.drshub.config.DrsProvider;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.regex.Pattern;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.ActiveProfiles;
 
 @SpringBootTest(classes = DrsHubApplication.class)
 @ActiveProfiles("human-readable-logging")
 public abstract class BaseTest {
+
+  @Autowired protected DrsHubConfig config;
+
+  protected ProviderHosts getProviderHosts(String provider) {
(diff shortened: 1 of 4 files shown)


service/src/test/java/bio/terra/drshub/BaseTest.java
```diff
@@ -1,19 +1,44 @@
...
+
   public void assertEmpty(Optional<?> optional) {
     assertTrue(optional.isEmpty(), "expected empty optional");
   }
 
   public void assertPresent(Optional<?> optional) {
     assertTrue(optional.isPresent(), "expected non-empty optional");
   }
+
+  protected record ProviderHosts(String drsUriHost, String dnsHost, DrsProvider drsProvider) {}
```
andy7i, line 43: looks like these were moved from `DrsHubApiControllerTest.java` -- was it moved here so that it could be used by other test suites?
(Also, it would've been nice to know this isn't new code!)

service/src/test/java/bio/terra/drshub/services/MetadataServiceTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class MetadataServiceTest extends BaseTest {
+
+  @Autowired private MetadataService metadataService;
+  @MockBean private DrsApi drsApi;
+
+  @MockBean private DrsApiFactory drsApiFactory;
+
+  @Test
+  void testDrsOptionsEndpoint() {
```
andy7i, line 23: Looks like this is testing 3 different cases. I think I've typically seen our unit tests organized so that they test specific behavior. So instead of testing 3 cases in one test, this would have been 3 unit tests to test out each "unit" of behavior. Any thoughts on that / why you decided to put all 3 in one test?

andy7i (approved):
couple questions but not blocking

service/src/test/java/bio/terra/drshub/BaseTest.java
tlangs: It was moved here so it can be re-used!

Shakespeared (approved):

service/src/test/java/bio/terra/drshub/services/MetadataServiceTest.java
tlangs: There are different ways of organizing unit tests. I like to organize tests by the method they're testing, since a lot of the time, there's mocking or setup that needs to be done. Replicating that setup over many tests make refactoring harder, especially if you need to update the tests for a method!

PR #45, opened 2022-08-26
ID-113 ID-115 ID-146 Use Discoverability endpoint authorizations

tlangs:
The Discoverability endpoint presents a problem for DRSHub. Currently, it relies on compile-time config to know how to authorize with a DRS Provider for an object. However, the discoverability feature in the new DRS Spec allows for runtime-evaluated authorization information on a per-object basis. So, DRSHub needs to to reconcile what it gets at runtime with what it has in it's config.

The way I've approached this is to have DRSHub generate functions that map a returned auth type to a configured auth type, and then that function is lazily evaluated once DRSHub actually knows the access type it's going to use. 

I've also gone though and done a java-ification of the code. Splitting things out into their own services, in my opinion, makes the code a bit more readable and modular, making future changes easier. I've also inverted some of the logic to follow more of a dispatcher pattern, rather than relying on null-checks. There's still some `nulls` there, but I avoided completely re-writing the logic.

Resolves
https://broadworkbench.atlassian.net/browse/ID-146
https://broadworkbench.atlassian.net/browse/ID-113
https://broadworkbench.atlassian.net/browse/ID-115


service/src/main/java/bio/terra/drshub/services/AuthService.java
```diff
@@ -0,0 +1,220 @@
...
+   * Build the authorizations that will enable the token/passport getting at runtime.
+   * DrsHubAuthorizations provide lazily-evaluated tokens based on access type.
+   *
+   * @param drsProvider Drs provider being reached out to
+   * @param components URI Components of the drs object
+   * @param bearerToken bearer token of the current user
+   * @return A list of DrsHubAuthorizations that can be used to get a token/passport based on object
+   *     info.
+   */
+  public List<DrsHubAuthorization> buildAuthorizations(
```
tlangs, line 72: Here's where the magic happens, where we build the authorizations that we will need to make objectInfo and accessUrl calls to a drs provider

service/src/main/java/bio/terra/drshub/services/AuthService.java
```diff
@@ -0,0 +1,220 @@
...
+import org.apache.commons.collections4.map.PassiveExpiringMap;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.HttpStatusCodeException;
+import org.springframework.web.client.RestClientException;
+import org.springframework.web.util.UriComponents;
+
+@Service
+@Slf4j
+public class AuthService {
```
tlangs, line 28: `AuthService` is a new service to encapsulate anything that talks to an auth provider, like bond or ECM.

service/src/main/java/bio/terra/drshub/config/DrsProviderInterface.java
```diff
@@ -122,14 +119,6 @@ && getAccessMethodConfigTypes().contains(AccessMethodConfigTypeEnum.gs)
         && Fields.overlap(requestedFields, Fields.BOND_SA_FIELDS);
   }
 
-  default boolean shouldFetchPassports(AccessMethod.TypeEnum accessMethodType) {
```
tlangs, line 125: We don't need this anymore because Drs Providers now tell us if they need a passport!

service/src/main/java/bio/terra/drshub/models/DrsHubAuthorization.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.function.Function;
+
+/**
+ * DrsHubAuthorization enables the mapping of an Authentication returned from a DRS Provider's
+ * Options endpoint to the auth defined in DrsHub's config. Since the returned Authorization is only
+ * available at runtime, but we have compile-time configs for each access type, which is only known
+ * after we get object info, the `getAuthForAccessType` function allows the runtime type to be used
+ * to get a lazily-evaluated bearer token or passport as needed.
+ */
+public record DrsHubAuthorization(
```
tlangs, line 16: This holds the lazily-evaluated function that actually gets the auth for a given provider and object.

service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
```diff
@@ -0,0 +1,99 @@
...
+
+@Slf4j
+@Service
+public record DrsProviderService(DrsHubConfig drsHubConfig) {
+
+  private static final Pattern compactIdRegex =
+      Pattern.compile("(?:dos|drs)://(?<compactId>dg\\.[0-9a-z-]+)", Pattern.CASE_INSENSITIVE);
+
+  private static final Pattern hostNameRegex =
+      Pattern.compile("(?:dos|drs)://(?<hostname>[^?/]+\\.[^?/]+)", Pattern.CASE_INSENSITIVE);
```
tlangs, line 22: This is the remediation of the catastrophic backtracking regex bug

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -0,0 +1,311 @@
...
+import java.util.Map;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestClientException;
+import org.springframework.web.util.UriComponents;
+
+@Service
+@Slf4j
+public record DrsResolutionService(
```
tlangs, line 32: This used to be `MetadataService`. Most helper methods here aren't actually new code, but appear as new because of the file rename.

service/src/test/resources/application-test.yaml
```diff
@@ -0,0 +1,160 @@
+deploy_env: ${DEPLOY_ENV:dev} # since there's a default, define this once
```
tlangs: This is a test config file that adds some providers to test specific cases that don't yet appear in production.

Shakespeared (approved):

service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
andy7i: Interestingly, I don't see this sonarcloud finding in recent scans: https://sonarcloud.io/summary/overall?id=DataBiosphere_terra-drs-hub

(the [old link](https://sonarcloud.io/project/security_hotspots?sinceLeakPeriod=true&status=FIXED&pullRequest=37&id=DataBiosphere_terra-drs-hub) no longer works)

issue: https://community.appway.com/screen/kb/article/checking-strings-avoiding-catastrophic-backtracking-1482810891360

github is not really showing a useful diff here. From what I can gather, this is the old query compared to the new query:

```
// old query
(?:dos|drs)://(?:(?<compactId>dg\\.[0-9a-z-]+)|(?<hostname>[^?/]+\\.[^?/]+))[:/](?<suffix>[^?]*)

// old query split into 2 queries(?):
(?:dos|drs)://(?:(?<compactId>dg\\.[0-9a-z-]+))[:/](?<suffix>[^?]*)
(?:dos|drs)://(?:(?<hostname>[^?/]+\\.[^?/]+))[:/](?<suffix>[^?]*)

// new/final queries
(?:dos|drs)://(?<compactId>dg\\.[0-9a-z-]+)
(?:dos|drs)://(?<hostname>[^?/]+\\.[^?/]+)
```

To save me time / reduce duplication in effort could you help me understand what was wrong with the old query and how your change fixes it?

service/src/test/resources/application-test.yaml
```diff
@@ -0,0 +1,160 @@
...
+
+drshub:
+  bondUrl: https://broad-bond-${deploy_env}.appspot.com
+  samUrl: https://sam.dsde-${deploy_env}.broadinstitute.org
+  externalcredsUrl: https://externalcreds.dsde-${deploy_env}.broadinstitute.org
+
+  compactIdHosts:
+    # it is important for case-insensitive matching that the keys here are all lower case
+    "dg.4503": gen3.biodatacatalyst.nhlbi.nih.gov
```
andy7i, line 11: this looks like a prod instance? do they have a staging env?

service/src/test/resources/application-test.yaml
```diff
@@ -0,0 +1,160 @@
...
+          must-revalidate: true
+        use-last-modified: false
+
+terra.common:
+  tracing:
+    stackdriverExportEnabled: ${CLOUD_TRACE_ENABLED:false}
+    samplingRate: ${SAMPLING_PROBABILITY:0}
+
+---
+spring.config.activate.on-profile: prod
```
andy7i: are these test configs for the prod environment? / do we have prod tests?

andy7i (commented):
partial review (I'm continuing to review in the meantime)

service/src/main/java/bio/terra/drshub/services/AuthService.java
```diff
@@ -0,0 +1,243 @@
...
+  private final BondApiFactory bondApiFactory;
+  private final DrsApiFactory drsApiFactory;
+  private final ExternalCredsApiFactory externalCredsApiFactory;
+
+  // To avoid absolutely hammering the ECM API during large batch analyses,
+  // cache the passport for a given bearer token for just a little bit.
+  // This also keeps DRSHub from calling ECM twice for the same request
+  // if the object info endpoint needs passport auth as well as the object access url endpoint.
+  private final Map<String, Optional<List<String>>> passportCache =
+      Collections.synchronizedMap(new PassiveExpiringMap<>(1, TimeUnit.MINUTES));
```
andy7i, line 39: this seems really useful! Should we get a security review for this since it does mean access could be out of date? It seems pretty minimal risk (being ~1 minute out of sync) but just in case that's important for Compliance.

tlangs: I'm pretty confident that this won't be an issue. Other apps cache PET accounts for a small amount of time without issue as an optimization.  Here's an example in TDR: https://github.com/DataBiosphere/jade-data-repo/blob/96deee3ac6964795d07a6106b8e0e8618da0be9b/src/test/java/bio/terra/common/auth/AuthService.java#L51-L58

service/src/test/resources/application-test.yaml
tlangs: They are prod urls, but they need to be for proper testing of URI validation. The tests don't actually reach out to them and mock all communication.

service/src/test/resources/application-test.yaml
tlangs: They are prod urls, but they need to be for proper testing of URI validation. The tests don't actually reach out to them and mock all communication.

tlangs:
@andy7i we marked the catastrophic backtracking issue as safe before, but it reared its head again here. https://sonarcloud.io/project/security_hotspots?id=DataBiosphere_terra-drs-hub&hotspots=AYHQKeywj3s7CzIsoSO_

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
andy7i: this is helpful! I skipped this file.

service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
andy7i: (We talked in person about my questions ^)

andy7i: What are your thoughts on pulling this out into a separate PR? I thought this change feels relatively atomic and a good candidate to be in its own commit in case we need to roll it back.

service/src/test/resources/application-test.yaml
andy7i: I think that might be worth documenting here since it's not intuitive and would break tests.

service/src/main/java/bio/terra/drshub/services/AuthService.java
andy7i: why are we only caching ECM and not drs or bond?

service/src/main/java/bio/terra/drshub/services/AuthService.java
```diff
@@ -0,0 +1,243 @@
...
+  /**
+   * Reach out to ECM and get the RAS Passport for the current user
+   *
+   * @param bearerToken token for the current user
+   * @return An Optional list of passports tied to the user.
+   */
+  public Optional<List<String>> fetchPassports(BearerToken bearerToken) {
+    return passportCache.computeIfAbsent(
+        bearerToken.getToken(),
+        token -> {
```
andy7i, line 226: this is cool! 

so the bearerToken get reused between requests?

also, would this be more accurate? (I'm not sure myself)
```suggestion
    return passportCache.computeIfAbsent(
        bearerToken.getToken(),
        passport -> {
```

tlangs: So, the arg here is still a token. Line 225 is the key to use to get an existing entry. `token` on line 226 is the token from the execution of `bearer.getToken()` if there's no matching entry, and is used as the key in the new entry created by the anonymous function.

service/src/main/java/bio/terra/drshub/services/AuthService.java
tlangs: We could, but in the spirit of minimal changes, we should do that in a separate PR.

service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
tlangs: I would like it to be a separate PR, but Sonarcloud flags it as a security issue and keeps me from merging the code, so its just kinda easier to do it here. We can easily change it back by just looking at the previous commit and making a hotfix.

service/src/test/resources/application-test.yaml
tlangs: Added a comment at the top of the file.

service/src/main/java/bio/terra/drshub/services/AuthService.java
andy7i: oh! that makes sense now!

service/src/main/java/bio/terra/drshub/services/AuthService.java
andy7i: agreed! was there any particular reason to start with caching ECM as opposed to others? Asking because I don't have any context around how often drs/bond/ecm get called.

service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
```diff
@@ -0,0 +1,99 @@
...
+            String.format(
+                "Could not find matching host for compact id [%s].",
+                compactIdMatch.group("compactId")));
+      }
+    } else if (hostNameMatch.find(0)) {
+      dnsHost = hostNameMatch.group("hostname");
+      strippedPath = drsUri.replaceAll("(?:dos|drs)://", "").replace(dnsHost + "/", "");
+    } else {
+      throw new BadRequestException(String.format("[%s] is not a valid DRS URI.", drsUri));
+    }
```
andy7i, line 65: why do we check for the compactId first and only check the hostName if there's no compactId? I don't see a reason for this ordering.

andy7i: found the answer, i think, which is that the order matters because compactIdMatch is a subset of hostNameMatch so we need to handle that case if it exists.

service/src/test/java/bio/terra/drshub/config/DrsProviderInterfaceTest.java
```diff
@@ -0,0 +1,91 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import bio.terra.drshub.BaseTest;
+import bio.terra.drshub.models.Fields;
+import bio.terra.drshub.services.DrsProviderService;
+import io.github.ga4gh.drs.model.AccessMethod;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+class DrsProviderInterfaceTest extends BaseTest {
```
andy7i, line 15: I don't think we're testing compactId URIs. I think we should have some tests to validate those.

service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
andy7i: not 100% sure but I think this might work / simplify the query?
```suggestion
  private static final Pattern compactIdRegex =
      Pattern.compile("(?:dos|drs)://(?<compactId>dg\\.[0-9a-z-]+)", Pattern.CASE_INSENSITIVE);

  private static final Pattern hostNameRegex =
      Pattern.compile("(?:dos|drs)://(?<hostname>[^?/]+)", Pattern.CASE_INSENSITIVE);
```

tlangs: Unfortunately, no. The removal of `\\.` means we would match invalid hosts. It removes the check for at least one period, so we would match `drs://foo/<object_id>`, which is invalid. Needs to be at least `drs://foo.bar/<object_id>`.

service/src/test/java/bio/terra/drshub/config/DrsProviderInterfaceTest.java
tlangs: The tests do cover compactIds. They are just unchanged and therefore don't show up in the diff. I did my best to not have to modify tests as a part of this change, as a way to verify that existing functionality didn't change.

andy7i: I see a couple tests that grab the compactId config values directly, but I don't see any compactId test coverage for `getUriComponents`. All of the tests I saw for `getUriComponents` use URIs. Are there other tests for this and/or `compactIdRegex`?

andy7i: To be explicit, I'm not asking for updates to existing tests. I think we should _add_ new tests to cover the compactIdRegex / code path.

tlangs: Ah, gotcha. Threw this together. Makes sure compactIds are mapped to their proper full hosts. The result from `getUriComponents` should never return the compactId, even if its supplied one! https://github.com/DataBiosphere/terra-drs-hub/blob/f042da9b7877f21a7ab1faca1eb816a5e94539ae/service/src/test/java/bio/terra/drshub/services/DrsProviderServiceTest.java#L33-L42

andy7i: thanks for adding this test!

service/src/test/java/bio/terra/drshub/services/DrsProviderServiceTest.java
```diff
@@ -28,4 +28,16 @@ void testResolvesdnsHostsAndProviders() {
...
     }
   }
+
+  @Test
+  void testMapsCompactIdsToTheirFullHosts() {
+    for (var entry : config.getCompactIdHosts().entrySet()) {
+      var compactId = entry.getKey();
+      var hostName = entry.getValue();
+      var testCompactUri = String.format("drs://%s/12345", compactId);
+      var expectedUri = String.format("//%s/12345", hostName);
```
andy7i, line 38: weird seeing a URI without a scheme/protocol in front! I looked through the code a bit but didn't figure out why it doesn't need the scheme in the URI 🤷‍♂️

andy7i (approved):
thanks for addressing much of the feedback and answering lots of questions! As my first meaty java PR, this was pretty difficult for me to grok so I appreciate your patience. I haven't groked all of it but I've been asked to thumb it and move on. I'm sure we'll improve our processes as we move forward as well and we don't need to make perfect the enemy of the good.

service/src/test/java/bio/terra/drshub/services/DrsProviderServiceTest.java
tlangs: Neither did I, but its what's already there!

PR #48, opened 2022-12-05
[ID-351] Enable signed urls for TDR 

Shakespeared:
Signed URLs are already turned on for Azure workspaces, but now that TDR is capable of serving signed URLs for data in Google requester pays buckets, we can enable it for Google as well. This one line change just flips the switch on that.

I also corrected a typo in one of the tests while I was in there. 

The Martha PR required test updates but this PR does not, as testing returning signed URLS is not linked to specific data hosts in DRSHub tests so the code path is already being exercised. 

Parallel PR for Martha here (merged): https://github.com/broadinstitute/martha/pull/264

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -100,6 +100,7 @@ jobs:
 
       - name: Secrets check
         run: |
+          sudo ln -s "$(which echo)" /usr/local/bin/say
           ./minnie-kenny.sh --force
           git secrets --scan-history
 
--- a/client-resttemplate/swagger.gradle
+++ b/client-resttemplate/swagger.gradle
@@ -1,6 +1,6 @@
 // All version controlled by dependency management plugin
 dependencies {
-	api 'org.springframework.boot:spring-boot-starter-json:+'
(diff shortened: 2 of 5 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -100,6 +100,7 @@ jobs:
 
       - name: Secrets check
         run: |
+          sudo ln -s "$(which echo)" /usr/local/bin/say
```
Shakespeared, line 103: copied from here https://github.com/broadinstitute/sam/pull/862, fixes an issue where git-secrets is trying to use the deprecated `say` command

jgainerdewar (approved):

client-resttemplate/swagger.gradle
```diff
@@ -1,6 +1,6 @@
 // All version controlled by dependency management plugin
 dependencies {
-	api 'org.springframework.boot:spring-boot-starter-json:+'
+	api 'org.springframework.boot:spring-boot-starter-json:2.7.6'
```
tlangs, line 3: Yes! Versions!

tlangs (approved):

PR #53, opened 2023-04-21
[ID-374] fix case insensitive matching for compact IDs 

Shakespeared:
This fixes a bug where the regex case insensitivity was not working correctly when matching DRS URI compact IDs, (for example dg.ANV0 would not match to dg.anv0). 

The regex patterns do have the `CASE_INSENSITIVE` flag already, however this did not seem to be behaving as expected. I suspect it may be something to do with the named capture groups. I wasn't able to make it work as is, so instead I am explicitly converting the URI string to lowercase before matching. 

Edited to add: realized that the problem was not the regex itself, it was the match to config values, which are all lowercase. So I do think that explicitly casting the compact ID to lowercase is the correct solution here.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
@@ -61,15 +61,17 @@ public record DrsProviderService(DrsHubConfig drsHubConfig) {
   public UriComponents getUriComponents(String drsUri) {
     UriComponents uriComponents;
 
-    var compactIdMatch = compactIdRegex.matcher(drsUri);
-    var hostNameMatch = hostNameRegex.matcher(drsUri);
+    // explicitly lowercase the string because the built-in case-insensitive regex matching is unreliable
+    var drsUriLowerCase = drsUri.toLowerCase();
+    var compactIdMatch = compactIdRegex.matcher(drsUriLowerCase);
+    var hostNameMatch = hostNameRegex.matcher(drsUriLowerCase);
 
     if (compactIdMatch.find(0)) {
       uriComponents = getCompactIdUriComponents(compactIdMatch);
     } else if (hostNameMatch.find(0)) {
       uriComponents = getHostnameUriComponents(hostNameMatch);
     } else {
-      throw new BadRequestException(String.format("[%s] is not a valid DRS URI.", drsUri));
+      throw new BadRequestException(String.format("[%s] is not a valid DRS URI.", drsUriLowerCase));
     }
 
     // Validate url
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
```diff
@@ -61,15 +61,17 @@ public record DrsProviderService(DrsHubConfig drsHubConfig) {
   public UriComponents getUriComponents(String drsUri) {
     UriComponents uriComponents;
 
-    var compactIdMatch = compactIdRegex.matcher(drsUri);
-    var hostNameMatch = hostNameRegex.matcher(drsUri);
+    // explicitly lowercase the uri because the case-insensitive regex flag is not working correctly
+    var drsUriLowerCase = drsUri.toLowerCase();
```
mbaumann-broad: Will this result in the DRS ID also being lowercased?
If so, that could potentially be an issue.
The [DRS (1.2+) specification](https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.2.0/docs/#section/DRS-IDs) allows uppercase characters in DRS IDs, and it whether the DRS server treats the DRS IDs as case sensitive or insensitive is an implementation detail left up to the DRS server.
Given this, it would be best to not case-shift the DRS ID.

As an alternate implementation, could just the compact identifier be lowercased in `getCompactIdUriComponents`?

mbaumann-broad (commented):
Thanks for the work on this @Shakespeared!

Please see the question/comment regarding case-shifting of the DRS ID.

service/src/main/java/bio/terra/drshub/services/DrsProviderService.java
Shakespeared: Yes it will, I was case-shifting the whole URI string to cover all my bases, but with the context that the IDs can be case-sensitive it make more sense to specifically shift the CIB only - I will update accordingly!

gpolumbo-broad (approved):

service/src/test/java/bio/terra/drshub/services/DrsProviderServiceTest.java
```diff
@@ -56,20 +56,33 @@ void testHostnameIdResolution() {
     }
   }
 
+  @Test
+  void testHostnameMatchingIsCaseInsensitive() {
+    var expectedUrl =
+        "drs://jade.datarepo-dev.broadinstitute.org/v1_e2151834-13cd-4156-9ea2-168a1b7abf60_0761203d-d2a1-448e-8f71-9f81d80ddd9d";
```
mbaumann-broad, line 62: Is the expectation that tests are only run in the dev tier?

If tests are also run in other tiers, the tier in the hostname should be parameterized.

mbaumann-broad (commented):
Looks good to me. Just one small question.

service/src/test/java/bio/terra/drshub/services/DrsProviderServiceTest.java
Shakespeared: Since this is a unit test and is only testing the compact identifier matching, I think it's ok to have this example URI be for dev tier only - however if you think it's better to test this in all tiers, I'm happy to do so!

mbaumann-broad: Thanks for your support @Shakespeared!
I am happy with this as it stands.

mbaumann-broad (approved):
Good to go! Thank you!

PR #54, opened 2023-05-02
Update configs for lower envs

Shakespeared:
ticket: https://broadworkbench.atlassian.net/browse/ID-546

This adds DRSHub config blocks to specify host URIs in specific lower environments (specifically alpha and staging). The configs should match the configs in Martha in functionality, although they are organized differently. It should map to the Martha configs (found [here](https://github.com/broadinstitute/martha/blob/dev/common/config.js)). 

The main change in this PR is to explicitly set the TDR URI (expanded from the new anvil CIB which is drs.anv0) to different values in dev, staging, and alpha. 

I went back and forth a bit on the best way to structure the config blocks, I think I landed on something that makes sense, but if this doesn't look right please let me know.

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -5,15 +5,6 @@ drshub:
   samUrl: https://sam.dsde-${deploy_env}.broadinstitute.org
   externalcredsUrl: https://externalcreds.dsde-${deploy_env}.broadinstitute.org
 
-  compactIdHosts:
-    # it is important for case-insensitive matching that the keys here are all lower case
-    # these are the CIBs we want in all environments, env specific ones are below
-    "dg.anv0": staging.theanvil.io
-    "drs.anv0": jade.datarepo-dev.broadinstitute.org
-    "dg.4dfc": nci-crdc-staging.datacommons.io
-    "dg.f82a1a": gen3staging.kidsfirstdrc.org
-    "dg.test0": ctds-test-env.planx-pla.net
-
   drsProviders:
     anvil:
       name: NHGRI Analysis Visualization and Informatics Lab-space (The AnVIL)
@@ -129,6 +120,7 @@ drshub:
     "dg.anv0": staging.theanvil.io # this is the old anvil CIB, the new one (drs.anv0) is set explicitly for each env
     "dg.4dfc": nci-crdc-staging.datacommons.io
     "dg.f82a1a": gen3staging.kidsfirstdrc.org
+    "dg.test0": ctds-test-env.planx-pla.net
 ---
 # prod
 spring.config.activate.on-profile: prod
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -102,16 +102,9 @@ terra.common:
     samplingRate: ${SAMPLING_PROBABILITY:0}
 
 ---
-spring.config.activate.on-profile: prod
-drshub:
-  compactIdHosts:
-    "dg.4503": gen3.biodatacatalyst.nhlbi.nih.gov
-    "dg.712c": staging.gen3.biodatacatalyst.nhlbi.nih.gov
-    "dg.anv0": gen3.theanvil.io
-    "drs.anv0": data.terra.bio
-    "dg.4dfc": nci-crdc.datacommons.io
-    "dg.f82a1a": data.kidsfirstdrc.org
----
+# not-prod
+# we want this test passport provider in all non-prod envs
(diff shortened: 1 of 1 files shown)


Ghost-in-a-Jar (approved):

service/src/main/resources/application.yml
```diff
@@ -132,4 +125,42 @@ drshub:
       mTlsConfig:
         keyPath: rendered/ras-mtls-client.key
         certPath: rendered/ras-mtls-client.crt
+  compactIdHosts:
+    "dg.4503": staging.gen3.biodatacatalyst.nhlbi.nih.gov
```
mbaumann-broad: `dg.4503` is the BDC compact id for production only.
It should never appear or be used in preprod tiers. 
Cross-deployment-tier data access does not work in general (for Gen3 or TDR)

If this line were deleted, and DRSHub unexpectedly received a DRS URI containing `dg.4503` would this result in the "compact identifier not found" error?
If so, then I think removing this line (or replacing it with a comment) is the appropriate change.

service/src/main/resources/application.yml
```diff
@@ -132,4 +125,42 @@ drshub:
...
+    "dg.4dfc": nci-crdc-staging.datacommons.io
+    "dg.f82a1a": gen3staging.kidsfirstdrc.org
+---
+# prod
+spring.config.activate.on-profile: prod
+drshub:
+  compactIdHosts:
+    "dg.4503": gen3.biodatacatalyst.nhlbi.nih.gov # biodatacatalyst
+    # this compact ID intentionally maps to the BDC staging url in prod
+    "dg.712c": staging.gen3.biodatacatalyst.nhlbi.nih.gov # biodatacatalyst
```
mbaumann-broad: `dg.712c` is the BDC compact id for preproduction only.
It should never appear or be used in the production tier. 
Cross-deployment-tier data access does not work in general (for Gen3 or TDR)

If this line were deleted, and DRSHub unexpectedly received a DRS URI containing `dg.712c` would this result in the "compact identifier not found" error?
If so, then I think removing this line (or replacing it with a comment) is the appropriate change.

mbaumann-broad (commented):
Thank you for these changes @Shakespeared!
Overall, the changes look good to me.
It will be great having DRSHub working as expected in the Alpha and Staging tiers, which currently include tests for Martha.

I left two comments regarding specific corner cases for BDC.
Please review these and post your thoughts. Thank you!

service/src/main/resources/application.yml
Shakespeared: Great info, thanks for surfacing this! You're correct in that if this is removed from specific tiers, it should throw the "compact identifier not found" error if a URL with this id is resolved. That sounds like the correct behavior to me, I'll update the configs

service/src/main/resources/application.yml
```diff
@@ -7,7 +7,6 @@ drshub:
 
   compactIdHosts:
     # it is important for case-insensitive matching that the keys here are all lower case
-    "dg.4503": gen3.biodatacatalyst.nhlbi.nih.gov
```
mbaumann-broad, line 10: The "prod" and "!prod" definitions for "dg.4503" and "dg.712c" later in this file look good and make sense.

I am less clear about when the definitions in this section would come into play.
Depending on when/how the values in this section are used, my general sense is that either:
- Both "dg.4503" and "dg.712c" would remain here.   
- or, both "dg.4503" and "dg.712c" would be removed here. 

Please advise. :)

Shakespeared: I had a look at the docs, the default case will appear in all envs unless explicitly overwritten, so I removed both of the env-specific CIBs from here, so each should only populate in the correct env now. Thanks for catching that!

mbaumann-broad: For the same reason, these can(/should) also be removed from the common block, as these all correctly appear in the `prod` and `!prod` blocks below:
```
    "dg.anv0": staging.theanvil.io
    "drs.anv0": jade.datarepo-dev.broadinstitute.org
    "dg.4dfc": nci-crdc-staging.datacommons.io
    "dg.f82a1a": gen3staging.kidsfirstdrc.org
  ```

mbaumann-broad (changes requested):
Thank you, @Shakespeared!
I think there is the opportunity for a little more cleanup along the same lines and for the same reasons.

mbaumann-broad (approved):
LGTM!
Thank you for the work on this, @Shakespeared!

PR #55, opened 2023-05-03
ID-377 Add types for openapi objects.

Ghost-in-a-Jar:
I am working on adding contract tests and noticed some fields could use more specific types. I regenerated the models locally and ran the tests.


Shakespeared (approved):
lgtm

tlangs (approved):

PR #57, opened 2023-05-09
ID-610 DRSHub gets signed URLs for GCS-backed DRS URIs

tlangs:
https://broadworkbench.atlassian.net/browse/ID-610

In order to retire Martha, DRSHub needs to replicate its functionality. DRSHub currently returns the Access URL defined in the [DRS Spec](https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.1.0/docs/#_accessurl), but won't actually return a signed URL for a GCS object using the fence account like Martha.

This PR replicates this functionality in DRSHub.


common/openapi.yml
```diff
@@ -20,6 +20,27 @@ paths:
           $ref: '#/components/responses/NotFound'
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/v4/gcs/getSignedUrl:
+    post:
+      summary: Get a signed URL for a GCS Object, signed the caller's fence service account
```
gpolumbo-broad:
```suggestion
      summary: Resolves the posted DRS URI and returns a signed URL for the GCS Object, signed by the caller's fence service account
```

What happens if the resolved DRS object does not have a gcs target?  For example, what if the canonical file is only stored on Azure and/or AWS?  Does a `404` returned by this endpoint mean the DRS object couldn't be found/resolved?  Or does it mean that the DRS URI was resolved but no gcs location was available to sign?

Wait, now I'm confused…looking at the request payload, this is really confusing.  I would have expected this just to be posted with a DRS URI not a project and bucket and stuff.  I'm going to read the rest of this PR to see if things make sense.  If you're reading this paragraph, it is because I forgot to come back to this comment and fix it based on what I learned reading the rest of the code

service/src/main/java/bio/terra/drshub/services/SignedUrlService.java
```diff
@@ -0,0 +1,54 @@
...
+    AuditLogger auditLogger) {
+
+  public URL getSignedUrl(
+      String bucket,
+      String objectName,
+      String dataObjectUri,
+      String googleProject,
+      BearerToken bearerToken,
+      String ip) {
+
```
gpolumbo-broad, line 36: I had to spend too much time making sure this is right in terms of business function and how this is expected to work.  I think you did it right in terms of taking the same args as Martha but this really doesn't make sense to me that it takes the `project`, `bucket`, and `objectName`.  That should be enough to uniquely identify an object.  But then it also takes a DRS URI, which seems duplicative at best and contradictory at worst.  For example, what is to stop me from using my DRS URI to authenticate, and then naming some _other_ project/bucket/object combo to try to get access to some _other_ object that isn't identified by the DRS URI?  Best case scenario, that is unintended functionality, worst case the user may be able to access files they weren't intended to have access to.  

Anyway, long story short, I guess making this API to match the same contract as Martha for now makes sense.  But I don't like it.  This just seems… not right?  Or not intuitive?  It can be improved :P

service/src/main/java/bio/terra/drshub/services/SignedUrlService.java
```diff
@@ -0,0 +1,54 @@
...
+    var components = drsProviderService.getUriComponents(dataObjectUri);
+    var drsProvider = drsProviderService.determineDrsProvider(components);
+    SaKeyObject saKey = authService.fetchUserServiceAccount(drsProvider, bearerToken);
```
gpolumbo-broad, line 39: If we are trying to mimic Martha's behavior, this needs to fall back to using the user's Pet SA if they are not authenticated with Bond.  https://github.com/broadinstitute/martha/blob/4e699ca160350b2702aef72201600e8ab91bd6c7/handlers/getSignedUrlV1.js#L12
https://github.com/broadinstitute/martha#get-signed-url

This may not be the ultimate behavior we want, but if we are just trying to port the Martha behavior to Bond, for now we should make it work this way.

service/src/main/java/bio/terra/drshub/services/SignedUrlService.java
```diff
@@ -0,0 +1,54 @@
...
+    Storage storage = googleStorageService.getAuthedStorage(saKey, googleProject);
+    BlobInfo blobInfo = BlobInfo.newBuilder(BlobId.of(bucket, objectName)).build();
+    var duration = drsHubConfig.getSignedUrlDuration();
+
+    var logEvent =
+        new AuditLogEvent.Builder()
+            .dRSUrl(dataObjectUri)
```
gpolumbo-broad, line 53: What about the `project/bucket/object`?   We're logging the DRS URI but not the bucket/object that we are _actually_ signing.

service/src/main/resources/application.yml
```diff
@@ -132,4 +132,6 @@ drshub:
       mTlsConfig:
         keyPath: rendered/ras-mtls-client.key
         certPath: rendered/ras-mtls-client.crt
+  signedUrlDuration: 1h
```
gpolumbo-broad, line 118: We should put this in the README and `openapi.yml` for this method so that users know how long the signed urls will work for.

gpolumbo-broad (changes requested):
Trevyn, I didn't finish going through all the files yet, but I feel like I had some important enough questions already in `SignedUrlService` that I figured I'd post this now and get this questions answered before proceeding.

service/src/main/java/bio/terra/drshub/services/SignedUrlService.java
tlangs: We need the drs object uri so that we can get the right fence account from Bond to sign the url. Without it, we have no clue what auth to use!

service/src/main/java/bio/terra/drshub/services/SignedUrlService.java
tlangs: We specifically decided against replicating the Martha behavior exactly in [this ticket](https://broadworkbench.atlassian.net/browse/ID-169). Sam signs URLs for gs paths not backed by DRS.

service/src/main/java/bio/terra/drshub/services/SignedUrlService.java
tlangs: We could do something different here and just provide the compact id or drs hostname. That would definitely work and be less redundant.

tlangs: Looked into this. We could make the bucket and object names optional, and if they don't exist, resolve the drs uri. But, we always need the drs uri, and I don't think its a good idea to encode the splitting of hostnames and compact IDs in the UI.

common/openapi.yml
tlangs: It does not resolve the DRS URI. It only uses it to get the right fence account key.

service/src/main/resources/application.yml
tlangs: Its in the OpenAPI docs now.

service/src/main/java/bio/terra/drshub/services/SignedUrlService.java
tlangs: The `AuditLogEvent` doesn't have a field for that, and I'm hesitant to add a field for fear of messing up some logging config downstream. The `drsUrl` should be enough for audit logging, since we can always resolve the object from that. Plus, the actual accessing of the gs url will be logged when the signed url is used.

tlangs:
I've also updated the github actions checkout version because of the Node12 deprecation. Figured that I should just do it while I'm in the repo.

Shakespeared (approved):

service/src/main/java/bio/terra/drshub/services/SignedUrlService.java
gpolumbo-broad: Yeah, I guess to me, Martha can do it all:

`Martha.getResolvedSignedUrl(drsUri)`

This thing can go resolve the object, determine the right provider, get the key from bond, and then sign the url returned by the resolution.  That seems the most obvious to me.  Either that or in any one of the other Martha calls where we resolve the DRS URI and give you back the key from Bond, just also return the signedUrl and leave it up to the client to decide which way they want to use to access the object.  Anyway, I guess I'm bringing us way off topic here, we won't be maintaining this stuff.  But I still have opinions 🙃

service/src/main/java/bio/terra/drshub/services/SignedUrlService.java
gpolumbo-broad: Well, not necessarily, we could send a DRS URI that doesn't point at the same project/bucket/object I think.  I don't recall seeing anything that enforced those matching data location of the DRS URI.

gpolumbo-broad (approved):
I'm holding off on getting over-critical of the tests in DRSHub for now.  I miss the descriptive `spec` naming in scalatest and I have thoughts about how to better organize and name and describe tests.  We _could_ try to set up some good standards now in this PR with the tests you wrote.  But I am also fine just leaving them as is and letting the owning team figure out their preferred way to rewrite and rename and reorganize tests.

PR #62, opened 2023-05-25
ID-377 Add pact can-i-deploy.

Ghost-in-a-Jar:


ichengchang:
@Ghost-in-a-Jar Could we copy the `logback-test.xml` from [https://github.com/broadinstitute/sam/blob/develop/pact4s/src/test/resources/logback.xml](https://github.com/broadinstitute/sam/blob/develop/pact4s/src/test/resources/logback.xml) (except [https://github.com/broadinstitute/sam/blob/develop/pact4s/src/test/resources/logback.xml#L7](https://github.com/broadinstitute/sam/blob/develop/pact4s/src/test/resources/logback.xml#L7)) to `service\src\test\resources` and rerun the verification test? It'll be interesting to see the provider logs echoed while the verification test is running. I think it's good for people to see the logs during the adoption phase. Later on we can reduce the level of logging if we want to.

ichengchang:
@Ghost-in-a-Jar  Does `AuditLogger` log to sentry? My question is could we confirm we are seeing the provider verification log where `AuditLogger` is supposed to log?

ichengchang (approved):
nit if you're able to see the verification log where `AuditLogger` is supposed to log. If not, we need to figure out and make it work.

service/analysis.gradle
```diff
@@ -43,6 +43,7 @@ test {
 }
 
 task verifyPacts(type: Test) {
+	outputs.upToDateWhen { false } // always run this task even if it's up to date
 	useJUnitPlatform {
 		includeTags 'Pact'
 	}
```
ichengchang, line 49:
```suggestion
	}
	testLogging {
		showStandardStreams = true
	}
```

ichengchang: By default, `JUnit` outputs the log to `xml` file. The config `showStandardStreams = true` will direct `logback` to the console so that the logs will show up in the verification workflow.

ichengchang (changes requested):
Added a suggestion on `testLogging`.

ichengchang (approved):
Looks great!

PR #64, opened 2023-06-26
[DR-3078] Pass through the google project to bill when requesting a signed URL

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3078

The Data Repo [getAccessURL](https://data.terra.bio/swagger-ui.html#/DataRepositoryService/GetAccessURL) and [postAccessURL](https://data.terra.bio/swagger-ui.html#/DataRepositoryService/PostAccessURL) endpoints have a "x-user-project" header that specifies which GCP project to bill when accessing the signed URL. 

**Changes**:
This PR passes in the `googleProject` from the DrsHub [getSignedUrl](https://drshub.dsde-dev.broadinstitute.org/#/gcs/getSignedUrl) endpoint as the "x-user-project" header when getting the access URL.

The [resolveDrs](https://drshub.dsde-dev.broadinstitute.org/#/drsHub/resolveDrs) endpoint can also fetch the access URL if certain conditions are met, but the google project is not specified in the post body. Would it be passed in as a header?

### DIFF
--- a/service/src/main/java/bio/terra/drshub/models/DrsApi.java
+++ b/service/src/main/java/bio/terra/drshub/models/DrsApi.java
@@ -13,4 +13,8 @@ public class DrsApi extends ObjectsApi {
   public void setBearerToken(String bearerToken) {
     ((OAuth) this.getApiClient().getAuthentication("BearerAuth")).setAccessToken(bearerToken);
   }
+
+  public void setHeader(String name, String value) {
+    this.getApiClient().addDefaultHeader(name, value);
+  }
 }
--- a/service/src/main/java/bio/terra/drshub/controllers/DrsHubApiController.java
+++ b/service/src/main/java/bio/terra/drshub/controllers/DrsHubApiController.java
@@ -39,12 +39,13 @@ public class DrsHubApiController implements DrsHubApi {
     var userAgent = request.getHeader("user-agent");
     var forceAccessUrl = Objects.equals(request.getHeader("drshub-force-access-url"), "true");
     var ip = request.getHeader("X-Forwarded-For");
+    var googleProject = request.getHeader("x-user-project");
 
     log.info("Received URL {} from agent {} on IP {}", body.getUrl(), userAgent, ip);
 
     var resourceMetadata =
         drsResolutionService.resolveDrsObject(
-            body.getUrl(), body.getFields(), bearerToken, forceAccessUrl, ip);
+            body.getUrl(), body.getFields(), bearerToken, forceAccessUrl, ip, googleProject);
 
     return ResponseEntity.ok(resourceMetadata);
   }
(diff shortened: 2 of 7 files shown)


service/src/main/java/bio/terra/drshub/models/DrsApi.java
```diff
@@ -13,4 +13,8 @@ public DrsApi(ApiClient apiClient) {
   public void setBearerToken(String bearerToken) {
     ((OAuth) this.getApiClient().getAuthentication("BearerAuth")).setAccessToken(bearerToken);
   }
+
+  public void setHeader(String name, String value) {
```
Shakespeared, line 17: do you mind adding a quick unit test to verify `setHeader`?

Shakespeared:
I'm not sure I fully understand the context of your question about `resolveDrs` fetching the access url, can you elaborate?

Shakespeared (changes requested):

samanehsan:
> I'm not sure I fully understand the context of your question about resolveDrs fetching the access url, can you elaborate?

Yeah! So when DRSHub gets a signed URL from TDR, it can specify the "x-user-project" header so that it can pass in the project to bill (for requester pays buckets). The `getSignedUrl` endpoint has a request body "googleProject" param, so I used that to set the header. But `resolveDrs`, which can also get a signed url, does not have a "googleProject" param. So I'm wondering if it's okay to have an "x-user-project" header for that endpoint that then gets passed into the "x-user-project" header for TDR.

service/src/main/java/bio/terra/drshub/models/DrsApi.java
samanehsan: I started writing this unit test but then I realized there's no way to get the `defaultHeaders` from the ApiClient (it's a private field and there's no getter). Do you have any suggestions on how to verify that the header was actually set in the test?

Shakespeared:
> > I'm not sure I fully understand the context of your question about resolveDrs fetching the access url, can you elaborate?
> 
> Yeah! So when DRSHub gets a signed URL from TDR, it can specify the "x-user-project" header so that it can pass in the project to bill (for requester pays buckets). The `getSignedUrl` endpoint has a request body "googleProject" param, so I used that to set the header. But `resolveDrs`, which can also get a signed url, does not have a "googleProject" param. So I'm wondering if it's okay to have an "x-user-project" header for that endpoint that then gets passed into the "x-user-project" header for TDR.

Thanks for clarifying! I think passing that through as a header is fine, that seems reasonable to me. You could probably add "googleProject" as an optional param to the `getSignedUrl` endpoint instead if that would be better, but I don't have an objection to just passing the header through as is.

service/src/main/java/bio/terra/drshub/models/DrsApi.java
Shakespeared: Hmm I took a look and I don't see a very easy way to do this, unless we want to add a getter for it. I think I'd be ok omitting this test as long as we're confident that it's tested on the TDR side.

Shakespeared (approved):

PR #66, opened 2023-07-12
ID-610 ID-667 Sam Signs Terra-Authed DRS Object Paths

tlangs:
DRSHub needs to sign URLs for DRS Objects using the `current_auth` authentication. We don't want to be passing around Service Account Keys anymore, and have made Sam to this for us. DRSHub needs to call out to Sam and return the response to the user.


sjkobori (approved):
Looks good to me

PR #68, opened 2023-07-28
[DR-3148] Support slashes in DRS Object Ids

nmalfroy:
This PR allows users to resolve drs object ids with slashes in them (this is a requirement of working with Azure drs ids).  E.g. this means that a user could call:
```
POST /api/v4/drs/resolve
{
  "url": "drs://jade.datarepo-dev.broadinstitute.org/foo/bar"
}
```
or
```
POST /api/v4/drs/resolve
{
  "url": "drs://jade.datarepo-dev.broadinstitute.org/foo%2Fbar"
}
```

and that will turn into DrsHub calling:
```
GET https://jade.datarepo-dev.broadinstitute.org/ga4gh/drs/v1/objects/foo%2Fbar
```

It's accomplished by updating how the client (only for ga4gh) is generated so that it eagerly encodes path/query parameters (even though the later isn't really needed, but forward proofing) and then doesn't re-encodes when generating the final URL to use to resolve DrsObjects.  It also eagerly.

It also attempts to decode object ids before sending them to be resolved.  I agree this is a strange thing but it lines up with how Martha works.

The testing is done via a modification to one of the existing integration tests and unit testing for the objectId decoding.

This is in conjunction with https://github.com/DataBiosphere/jade-data-repo/pull/1485 and brings this behavior to parity with Martha

Review note: commit-by-commit will make this much easier to review (e.g. I isolated the base template introduction commit and the changes).


integration/src/main/java/scripts/testscripts/ResolveTdrDrs.java
```diff
@@ -6,38 +6,48 @@
...
 public class ResolveTdrDrs extends TestScript {
-  private static String TDR_TEST_DRS_URI =
-      "drs://jade.datarepo-dev.broadinstitute.org/v1_93dc1e76-8f1c-4949-8f9b-07a087f3ce7b_8b07563a-542f-4b5c-9e00-e8fe6b1861de";
+  // All URIs pont to the same TDR hosted DRS object. Some URIs are aliases created in the TDR
+  // instance
+  private static final List<String> TDR_TEST_DRS_URIS =
+      List.of(
+          "drs://jade.datarepo-dev.broadinstitute.org/v1_93dc1e76-8f1c-4949-8f9b-07a087f3ce7b_8b07563a-542f-4b5c-9e00-e8fe6b1861de",
+          "drs://jade.datarepo-dev.broadinstitute.org/drshubfoo",
+          "drs://jade.datarepo-dev.broadinstitute.org/drshubfoo/bar");
```
tlangs: Did we want to add a url containing `%2F` here as a test? Also, we should talk about Pact testing between DRSHub and TDR!

tlangs (changes requested):
Could we unit-test the generated code? The generated `ApiClient` should have an `ApiClient(RestTemplate restTemplate)` constructor, and we could pass a mock `RestTemplate` into that and make sure that the URL Encoding is happening as expected. Would also let us test against new versions of the DRS API spec and swagger codegen upgrades.

integration/src/main/java/scripts/testscripts/ResolveTdrDrs.java
nmalfroy: Ah good catch.  I'll add that

nmalfroy:
I was trying to find a way to test it given the existing framework and it didn't seem like a thing that was practical (e.g. we need a handle on the low level HTTP client and I don't think the client is factored in a way that lets us get that).  That's why I went the integration test route to test that it does the right thing but it's an interesting idea to try with the resttemplate mocking. I'll see what I can do

nmalfroy:
Ok @tlangs I updated the unit tests I added to use a mocked out rest template (I left the existing unit tests alone since mocking that out would get real unwieldy)

tlangs (approved):

PR #69, opened 2023-08-02
ID-725 Cache Bond Responses

tlangs:
DRSHub makes 1 or 2 requests to Bond for each DRS URL Resolution, depending on what fields are requested. This means that DRSHub makes a huge amount of requests for a single user’s workflow. If we cache Bond responses by Bearer Token and DRS Provider, we should be able to drastically cut down on the number of requests.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/AuthService.java
+++ b/service/src/main/java/bio/terra/drshub/services/AuthService.java
@@ -41,6 +41,14 @@ public class AuthService {
   private final Map<String, Optional<List<String>>> passportCache =
       Collections.synchronizedMap(new PassiveExpiringMap<>(1, TimeUnit.MINUTES));
 
+  // For every DRS Resolution requiring a signed URL using Bond authorization,
+  // we need to reach out to Bond twice:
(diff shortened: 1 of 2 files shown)


Shakespeared (approved):

service/src/main/java/bio/terra/drshub/services/AuthService.java
```diff
@@ -40,6 +41,12 @@ public class AuthService {
   private final Map<String, Optional<List<String>>> passportCache =
       Collections.synchronizedMap(new PassiveExpiringMap<>(1, TimeUnit.MINUTES));
 
+  private final Map<Pair<String, String>, SaKeyObject> serviceAccountKeyCache =
```
Ghost-in-a-Jar, line 53: can you comment what the key pair is here just for posterity

Ghost-in-a-Jar (approved):

service/src/main/java/bio/terra/drshub/services/AuthService.java
tlangs: Sure can!

PR #75, opened 2023-09-11
DR-3214 Set Headers per-request for DRS API Clients

tlangs:
We need to use different `DrsApi` objects because, if we don't, all calls to the same provider will share headers.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
@@ -3,15 +3,11 @@ package bio.terra.drshub.services;
 import bio.terra.drshub.config.DrsProvider;
 import bio.terra.drshub.models.DrsApi;
 import io.github.ga4gh.drs.client.ApiClient;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Objects;
 import lombok.extern.slf4j.Slf4j;
 import nl.altindag.ssl.SSLFactory;
 import nl.altindag.ssl.util.Apache4SslUtils;
 import nl.altindag.ssl.util.PemUtils;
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
```diff
@@ -28,33 +24,25 @@
...
-                  ? new ApiClient(makeRestTemplateWithPooling())
-                  : new ApiClient(
-                      makeMTlsRestTemplateWithPooling(
-                          mTlsConfig.getCertPath(), mTlsConfig.getKeyPath()));
+    log.info(
+        "Creating new DrsApi client for host '{}', for DRS Provider '{}'",
+        uriComponents.getHost(),
+        drsProvider.getName());
+    var mTlsConfig = drsProvider.getMTlsConfig();
+    var drsClient =
```
nmalfroy, line 34: Are we worried about overhead from getting rid of the cache?  I guess we can up replicas if needed?

nmalfroy (approved):
Looks good. We'll need to keep an eye on how it scales (we could always reintroduce the caching by updating the client to allow dynamically sending headers

service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
tlangs: I have concerns too, but this was more of a premature optimization that didn't pan out. The most latency was coming from requests to Bond, and the caches in front of that should be doing the bulk of the speedup. The only thing this would do is cause more heap space to be needed, but we can tune that pretty easily.

nmalfroy: Yeah, that sounds totally reasonable

sjkobori (approved):

PR #76, opened 2023-09-19
[DR-3077] Log DRS resolution and URL signing to Bard

nmalfroy:
This PR:
- Adds a Bard client definition and client to send events to Bard for it to track all successful drs resolution calls.  The `EventProperties` is interesting since it's a combination of set fields and random extra values.  I needed to add a custom serializer to properly handle that
- Adds a request interceptor to asynchronously send these events (these are non-blocking so there should be no perf impact other that this will add Runnables to the async queue).  Only methods tagged with the newly created `@TrackEvent` annotation will be tracked.
- Worth calling out again: only successful drs resolutions will be logged since this is to track what files folks are using.  We should use the usual methods (e.g. grafana) for logging excessive 404s for instance
- It's configurable but by default, the flag is set to NOT send events to Mixpanel when they reach Bard so the requests will be logged in the Warehouse.

At the end of the day, this ends up sending events to Bard that look like:
```json
{
  "event": "drshub:api",
  "properties": {
    "appId": "drshub",
    "pushToMixpanel": false,
    "statusCode": 200,
    "requestUrl": "/api/v4/drs/resolve or /api/v4/gcs/getSignedUrl",
    "userProject": "<the x-user-project header value>",
    "ip": "<the X-Forwarded-For header value>",
    "forceAccessUrl": "<the drshub-force-access-url header value>"
  }
}
```

where the last three are only sent if there is a header value set. This should give us enough information (the `userProject` field in particular)

This is better reviewed commit-by-commit

### DIFF
--- a/service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
+++ b/service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
@@ -9,6 +9,8 @@ import org.immutables.value.Value;
 @PropertiesInterfaceStyle
 public interface DrsHubConfigInterface {
 
+  String getBardUrl();
+
   String getBondUrl();
 
   String getSamUrl();
@@ -30,4 +32,8 @@ public interface DrsHubConfigInterface {
   Integer getPencilsDownSeconds();
 
   Integer asyncThreads();
+
+  // If this is true, then we will track calls to the Bard API in MixPanel in addition to the
+  // BigQuery Data warehouse.
+  Boolean trackInMixPanel();
 }
--- a/service/src/main/java/bio/terra/drshub/controllers/DrsHubApiController.java
+++ b/service/src/main/java/bio/terra/drshub/controllers/DrsHubApiController.java
@@ -7,6 +7,7 @@ import bio.terra.drshub.generated.model.RequestObject;
 import bio.terra.drshub.generated.model.ResourceMetadata;
 import bio.terra.drshub.models.Fields;
 import bio.terra.drshub.services.DrsResolutionService;
+import bio.terra.drshub.tracking.TrackCall;
 import bio.terra.drshub.util.AsyncUtils;
 import java.util.ArrayList;
 import java.util.Objects;
@@ -25,6 +26,7 @@ public record DrsHubApiController(
     implements DrsHubApi {
 
   @Override
+  @TrackCall
   public ResponseEntity<ResourceMetadata> resolveDrs(RequestObject body) {
     var bearerToken = bearerTokenFactory.from(request);
     validateRequest(body);
--- a/service/src/main/java/bio/terra/drshub/services/BardApiFactory.java
+++ b/service/src/main/java/bio/terra/drshub/services/BardApiFactory.java
@@ -1,0 +1,94 @@
+package bio.terra.drshub.services;
+
+import bio.terra.bard.api.BardApi;
+import bio.terra.bard.client.ApiClient;
+import bio.terra.bard.model.EventProperties;
+import bio.terra.common.iam.BearerToken;
+import bio.terra.drshub.DrsHubException;
+import bio.terra.drshub.config.DrsHubConfig;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.util.List;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestTemplate;
+
+@Service
+public record BardApiFactory(DrsHubConfig drsHubConfig) {
+
+  public BardApi getApi(BearerToken bearerToken) {
+    var bardApi = new BardApi(new BardApiClient());
+
+    bardApi.getApiClient().setBasePath(drsHubConfig.getBardUrl());
+    bardApi.getApiClient().setAccessToken(bearerToken.getToken());
+
+    return bardApi;
+  }
+
+  /**
+   * Extension of the BardApiClient that overrides the buildRestTemplate method to add a custom
+   * serializer for EventProperties
+   */
+  private static class BardApiClient extends ApiClient {
+    @Override
+    protected RestTemplate buildRestTemplate() {
+      RestTemplate restTemplate = super.buildRestTemplate();
+
+      MappingJackson2HttpMessageConverter converter = new MappingJackson2HttpMessageConverter();
+      converter.setObjectMapper(configureObjectMapper());
+      restTemplate.setMessageConverters(List.of(converter));
+      return restTemplate;
+    }
+  }
+
+  @VisibleForTesting
+  static ObjectMapper configureObjectMapper() {
+    ObjectMapper objectMapper = new ObjectMapper();
+    SimpleModule module = new SimpleModule();
+    module.addSerializer(EventProperties.class, new EventPropertiesSerializer());
+    objectMapper.registerModule(module);
+    return objectMapper;
+  }
+
+  /**
+   * Custom serializer for EventProperties. This is needed because the Bard API expects the
+   * arbitrary user properties to be serialized as top-level fields in the JSON object along with
+   * the first class fields (e.g. appId and pushToMixpanel)
+   */
+  private static class EventPropertiesSerializer extends StdSerializer<EventProperties> {
+    private static final List<String> RESERVED_FIELDS = List.of("appId", "pushToMixpanel");
+
+    public EventPropertiesSerializer() {
+      this(null);
+    }
+
+    public EventPropertiesSerializer(Class<EventProperties> t) {
+      super(t);
+    }
+
+    @Override
+    public void serialize(EventProperties value, JsonGenerator jgen, SerializerProvider provider)
+        throws IOException {
+      jgen.writeStartObject();
+      jgen.writeStringField("appId", value.getAppId());
+      jgen.writeBooleanField("pushToMixpanel", value.isPushToMixpanel());
+      value.forEach(
+          (key, val) -> {
+            try {
+              // Ignore reserved fields since they are already written above
+              if (!RESERVED_FIELDS.contains(key)) {
+                jgen.writeObjectField(key, val);
+              }
+            } catch (IOException e) {
+              throw new DrsHubException("Error serializing event properties", e);
+            }
+          });
+      jgen.writeEndObject();
+    }
+  }
+}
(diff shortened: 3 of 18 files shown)


service/src/main/resources/vendor/bard.yaml
okotsopoulos, line 1: Where did this definition come from?  If Bard changes, I'm guessing this will be out of date, could that pose an issue?

service/src/main/resources/vendor/bard.yaml
```diff
@@ -0,0 +1,110 @@
...
+        # See bio.terra.drshub.services.BardApiFactory.EventPropertiesSerializer
+        appId:
+          type: string
+        pushToMixpanel:
+          type: boolean
+      additionalProperties:
+        type: object
+      required:
+        - appId
+    Itentify:
```
okotsopoulos: That's a funny name :) do you know why this is `Itentify` instead of `Identify`, is the latter a reserved word?

service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
```diff
@@ -30,4 +32,8 @@ public interface DrsHubConfigInterface {
   Integer getPencilsDownSeconds();
 
   Integer asyncThreads();
+
+  // If this is true, then we will track calls to the Bard API in MixPanel in addition to the
```
okotsopoulos, line 36: Nit: this might be nice as a docstring instead, I think that might make the comment more visible when viewing references in IDEs.

service/src/main/java/bio/terra/drshub/services/BardApiFactory.java
```diff
@@ -0,0 +1,94 @@
...
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.util.List;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestTemplate;
+
+@Service
+public record BardApiFactory(DrsHubConfig drsHubConfig) {
```
okotsopoulos, line 22: Is this a place where we'd benefit from sharing common HTTP clients?  I wasn't 100% sure if it was relevant here.  We made that switch in TDR and it helped our memory usage quite a bit:

https://github.com/DataBiosphere/jade-data-repo/pull/1462

service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
okotsopoulos: Also, is there a reason that we've named this slightly off from the existing convention of `pushToMixpanel`?

service/src/main/java/bio/terra/drshub/controllers/DrsHubApiController.java
```diff
@@ -25,6 +26,7 @@ public record DrsHubApiController(
     implements DrsHubApi {
 
   @Override
+  @TrackCall
```
okotsopoulos, line 29: This is neat!

service/src/main/java/bio/terra/drshub/tracking/TrackCall.java
```diff
@@ -0,0 +1,14 @@
+package bio.terra.drshub.tracking;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Add this annotation to a method to track its call as a user event in Bard. Note this should just
+ * be used within Controller classes
```
okotsopoulos: Why should this be used within Controller classes only, is it to guard against multiple logs for the same root event?

service/src/main/java/bio/terra/drshub/services/TrackingService.java
```diff
@@ -0,0 +1,71 @@
...
+    var bardApi = bardApiFactory.getApi(bearerToken);
+    // Sync the user profile if needed.
+    syncUser(bardApi, bearerToken);
+    // Log the user event
+    logEvent(bardApi, eventName, properties);
+  }
+
+  /**
+   * Call the syncProfile endpoint in Bard. This ensures that the Terra user is properly associated
+   * with a MixPanel user. Note that this is if mixpanel tracking is enabled.
```
okotsopoulos, line 45: I'm not solid here on the behavior when mixpanel tracking isn't enabled: is this method never called, do the internals behave as no-ops, etc?

service/src/main/java/bio/terra/drshub/tracking/TrackingInterceptor.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @Override
+  public void afterCompletion(
+      HttpServletRequest request,
+      @NotNull HttpServletResponse response,
+      @NotNull Object handler,
+      Exception ex) {
+    var path = request.getRequestURI();
+    var responseStatus = HttpStatus.valueOf(response.getStatus());
+    // Note: invalid responses will not be logged since there are no sessions to associate users
```
okotsopoulos, line 46: I appreciate this comment!

okotsopoulos (commented):
I am very close to being done with this review, just want to look more closely at the `TrackingServiceTest` and `TrackingInterceptorTest` :) here are my initial questions!

service/src/main/java/bio/terra/drshub/tracking/TrackingInterceptor.java
```diff
@@ -0,0 +1,103 @@
...
+      var properties =
+          new HashMap<String, Object>(
+              Map.of("statusCode", responseStatus.value(), "requestUrl", path));
+
+      properties.putAll(readBody(request));
+
+      addToPropertiesIfPresentInHeader(request, properties, "x-user-project", "userProject");
+      addToPropertiesIfPresentInHeader(request, properties, "X-Forwarded-For", "ip");
+      addToPropertiesIfPresentInHeader(
+          request, properties, "drshub-force-access-url", "forceAccessUrl");
```
okotsopoulos, line 62: I am missing the intention behind pulling these specific header values into properties, but I think comment(s) would help (or breaking the code out further).

okotsopoulos (approved):
Looks good!  No critical blockers in my comments.

service/src/main/java/bio/terra/drshub/tracking/TrackingInterceptor.java
nmalfroy: Oh that's because those are the only known headers (that I could find) and I didn't want to just pass along all headers since that could include some sensitive info (like auth tokens, etc.)

service/src/main/java/bio/terra/drshub/services/TrackingService.java
nmalfroy: Oh that's branching that lives in Bard.  We always log to the warehouse but we can opt out of _also_ logging to mixpanel (which we want to do in this case in all likelihood to not explode our quota)

service/src/main/java/bio/terra/drshub/tracking/TrackCall.java
nmalfroy: Yeah, I wanted to log the explicit controller call since it maps nicely to a URL and HTTP method which is ultimately what we're interested (no value in tracking internal method calls)

service/src/main/java/bio/terra/drshub/controllers/DrsHubApiController.java
nmalfroy: I thought this was a cool approach :)

service/src/main/java/bio/terra/drshub/services/BardApiFactory.java
nmalfroy: Hmmm maybe.  Let me look

service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
nmalfroy: Nope.  I'll update!

service/src/main/resources/vendor/bard.yaml
nmalfroy: That's because I derped :). I'll update

service/src/main/resources/vendor/bard.yaml
nmalfroy: Yeah, I started from the swagger that gets generated from the api definition there. I ended up needing to do quite a bit of hand building to make it play nice with the code gen.  I think for now it's ok but ultimately, yes we'll want some way to keep this in sync (thinking pact tests)

service/src/main/java/bio/terra/drshub/tracking/TrackingInterceptor.java
okotsopoulos: Got it!  I think this would help grok-ability long-term to have in a code comment or something like that.

service/src/main/java/bio/terra/drshub/tracking/TrackCall.java
okotsopoulos: Cool!  That could be helpful context to add in the docstring.

nmalfroy: fair enough!

service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
tlangs: Is this ever going to be `true`?

service/src/main/java/bio/terra/drshub/services/BardApiFactory.java
tlangs: It can be treacherous to share api clients: https://github.com/DataBiosphere/terra-drs-hub/pull/75

tlangs: Especially since each API Client has the bearer-token baked into it, and it can't be set per-request.

service/src/main/java/bio/terra/drshub/services/TrackingService.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private final Map<String, String> bearerTokenCache =
+      Collections.synchronizedMap(new PassiveExpiringMap<>(15, TimeUnit.MINUTES));
+
+  public TrackingService(BardApiFactory bardApiFactory, DrsHubConfig config) {
+    this.bardApiFactory = bardApiFactory;
+    this.trackInMixpanel = config.trackInMixPanel();
+  }
+
+  @Async("asyncExecutor")
```
tlangs, line 34: Is using the same executor as the requests going to cut into how performant DRSHub is?

service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
nmalfroy: Likely no, but it made it easier to test so I left it in as a potential knob

service/src/main/java/bio/terra/drshub/services/BardApiFactory.java
nmalfroy: Yeah, I'll leave it separate

service/src/main/java/bio/terra/drshub/services/TrackingService.java
nmalfroy: I was wondering about that.  I kinda doubt it since it's already set to Tomcat max threads.  One more reason to bump the JDK and us virtual threads

tlangs (approved):

PR #77, opened 2023-09-27
ID-822 Use new Sam endpoint for Signed URLs

tlangs:
A new Signed URL Endpoint is being created in Sam to correctly handle signed urls for requester-pays buckets. DRSHub should use it.

https://github.com/broadinstitute/sam/pull/1193

### DIFF
--- a/service/src/main/resources/vendor/sam.yaml
+++ b/service/src/main/resources/vendor/sam.yaml
@@ -20,7 +20,7 @@ servers:
 paths:
   /api/google/v1/user/signedUrlForBlob:
     post:
-      summary: gets a signed url for a blob using the arbitrary pet service account for the user and an optional requester pays google project
+      summary: gets a signed url for a blob using an optional requester pays google project
       tags: [ sam ]
       operationId: signedUrlForBlob
       requestBody:
@@ -30,8 +30,7 @@ paths:
             schema:
               type: object
               required:
-                - bucketName
-                - blobName
+                - gsPath
               properties:
                 gsPath:
                   type: string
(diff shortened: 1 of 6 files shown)


tlangs:
Integration test is expected to fail until https://github.com/broadinstitute/sam/pull/1193 ships

service/src/main/resources/vendor/sam.yaml
```diff
@@ -18,19 +18,11 @@ servers:
     description: Development
 
 paths:
-  /api/google/v1/user/petServiceAccount/{project}/signedUrlForBlob:
+  /api/google/v1/user/signedUrlForBlob:
     post:
-      summary: gets a signed url for a blob using the pet service account for the user in the provided project
+      summary: gets a signed url for a blob using the arbitrary pet service account for the user and an optional requester pays google project
```
nmalfroy: Again, I'd love to understand why it needs to be arbitrary.  The project should get passed in by the client.  I worry this will cause permissions issues

nmalfroy (approved):
Same comment about the UI PR: I have my doubts about the API

Shakespeared (approved):

service/src/test/java/bio/terra/drshub/util/SignedUrlTestUtils.java
```diff
@@ -95,7 +95,7 @@ public static void setupDrsResolutionServiceMocks(
             any(BearerToken.class),
             eq(true),
             nullable(String.class),
-            nullable(String.class));
+            eq(googleProject));
```
nmalfroy: I'm like 90% sure this is causing the tests to fail

tlangs: It totes was.

service/src/main/resources/vendor/sam.yaml
tlangs: Removed all references to Pet Service Accounts

PR #80, opened 2023-11-02
[DR-3324] Only fetch DRS object via passport if one exists

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3324

`DrsHub` asks TDR Production for the supported authorization mechanisms for a controlled access GTEx file:
```
curl -X 'OPTIONS' \
  'https://data.terra.bio/ga4gh/drs/v1/objects/v2_e559f418-7452-31c2-94fe-ed412157363d' \
  -H 'accept: application/json'

{
  "supported_types": [
    "PassportAuth",
    "BearerAuth"
  ],
  "passport_auth_issuers": [
    "https://stsstg.nih.gov/"
  ],
  "bearer_auth_issuers": null
}
```
Seeing that PassportAuth is a supported type (dictated by the presence of both a PHS ID and consent code on the source snapshot), it tries to obtain a passport to use for fetching the DRS object.

**Previous behavior**

In cases where DRS OPTIONS indicate `PassportAuth` as a supported means of authorization, if a user Passport is not available, then DrsHub passed an empty string as a Passport when attempting object resolution.

But the empty string is not a valid Passport: when DrsHub passed it to TDR, TDR passed it to the ECM client for decoding.  ECM then throws an unhandled error:
```
{
  "message" : "500 Internal Server Error: \"{\"msg\":\"400 Bad Request: \\\"{<EOL>  \\\"message\\\" : \\\"invalid passport jwt\\\",<EOL>  \\\"statusCode\\\" : 400<EOL>}\\\"\",\"status_code\":500}\"",
  "statusCode" : 500
}
```

Furthermore, DrsHub did not fall back to fetching the object via bearer token.

**New behavior**

In cases where DRS OPTIONS indicate `PassportAuth` as a supported means of authorization, DrsHub only attempts to fetch the object via POSTed passport if such a passport exists.

If any part of this fails, DrsHub now falls back to fetching the object via bearer token.

**Tests**

I added targeted unit tests covering DRS object fetching, but there were no previous unit tests for `DrsResolutionService`: I believe this code remains largely uncovered by automated tests, but am also still finding my way around this repo.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -220,11 +220,10 @@ public class DrsResolutionService {
     var sendMetadataAuth = drsProvider.isMetadataAuth();
 
     var objectId = getObjectId(uriComponents);
-    log.info(
-        "Requesting DRS metadata for {} with auth required {} from host {}",
-        drsUri,
-        sendMetadataAuth,
-        uriComponents.getHost());
+    String drsRequestLogMessage =
+        "Requesting DRS metadata for %s with auth required %s from host %s"
+            .formatted(drsUri, sendMetadataAuth, uriComponents.getHost());
+    log.info(drsRequestLogMessage);
 
     var drsApi = drsApiFactory.getApiFromUriComponents(uriComponents, drsProvider);
     if (sendMetadataAuth) {
@@ -233,8 +232,16 @@ public class DrsResolutionService {
       drsApi.setBearerToken(bearerToken.getToken());
       if (authorizations.stream()
           .anyMatch(a -> a.drsAuthType() == Authorizations.SupportedTypesEnum.PASSPORTAUTH)) {
-        Optional<List<String>> passports = authService.fetchPassports(bearerToken);
-        return drsApi.postObject(Map.of("passports", passports.orElse(List.of(""))), objectId);
+        List<String> passports = authService.fetchPassports(bearerToken).orElse(List.of());
+        if (!passports.isEmpty()) {
+          try {
+            return drsApi.postObject(Map.of("passports", passports), objectId);
+          } catch (Exception ex) {
+            // We are catching a general exception to ensure that we fall back to getting the object
+            // via bearer token in case of any failure
+            log.warn(drsRequestLogMessage + " failed via passport, using bearer token", ex);
+          }
+        }
       }
     }
 
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -234,15 +234,15 @@ public class DrsResolutionService {
       drsApi.setBearerToken(bearerToken.getToken());
       if (authorizations.stream()
           .anyMatch(a -> a.drsAuthType() == Authorizations.SupportedTypesEnum.PASSPORTAUTH)) {
-        List<String> passports = authService.fetchPassports(bearerToken).orElse(List.of());
-        if (!passports.isEmpty()) {
-          try {
+        try {
+          List<String> passports = authService.fetchPassports(bearerToken).orElse(List.of());
+          if (!passports.isEmpty()) {
             return drsApi.postObject(Map.of("passports", passports), objectId);
-          } catch (Exception ex) {
-            // We are catching a general exception to ensure that we fall back to getting the object
-            // via bearer token in case of any failure
-            log.warn(drsRequestLogMessage + " failed via passport, using bearer token", ex);
           }
+        } catch (Exception ex) {
+          // We are catching a general exception to ensure that we fall back to getting the object
+          // via bearer token in case of any failure
+          log.warn(drsRequestLogMessage + " failed via passport, using bearer token", ex);
         }
       }
     }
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -233,8 +234,16 @@ private DrsObject fetchObjectInfo(
       drsApi.setBearerToken(bearerToken.getToken());
       if (authorizations.stream()
           .anyMatch(a -> a.drsAuthType() == Authorizations.SupportedTypesEnum.PASSPORTAUTH)) {
-        Optional<List<String>> passports = authService.fetchPassports(bearerToken);
-        return drsApi.postObject(Map.of("passports", passports.orElse(List.of(""))), objectId);
+        List<String> passports = authService.fetchPassports(bearerToken).orElse(List.of());
+        if (!passports.isEmpty()) {
```
pshapiro4broad: IMO this is exactly the reason why an API that returns a collection shouldn't use Optional. The calling code needs to check for both the optional being present _and_ the collection being non-empty. Returning an empty collection instead would make this code simpler. But it looks like `Optional` is threaded through many places here so it wouldn't be easy to change this.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -233,8 +234,16 @@ private DrsObject fetchObjectInfo(
       drsApi.setBearerToken(bearerToken.getToken());
       if (authorizations.stream()
           .anyMatch(a -> a.drsAuthType() == Authorizations.SupportedTypesEnum.PASSPORTAUTH)) {
-        Optional<List<String>> passports = authService.fetchPassports(bearerToken);
-        return drsApi.postObject(Map.of("passports", passports.orElse(List.of(""))), objectId);
+        try {
+          List<String> passports = authService.fetchPassports(bearerToken).orElse(List.of());
+          if (!passports.isEmpty()) {
+            return drsApi.postObject(Map.of("passports", passports), objectId);
```
nmalfroy, line 240: Too bad...I wish that the code gen would make postObject take a typed object. I'mm actually kinda surprised that it doesn't

nmalfroy (approved):
This looks great!  And thanks for the general doc cleanup too!

service/src/test/java/bio/terra/drshub/services/DrsResolutionServiceTest.java
```diff
@@ -0,0 +1,207 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.web.client.RestClientException;
+import org.springframework.web.util.UriComponents;
+
+@Tag("Unit")
+@ContextConfiguration(classes = {DrsResolutionService.class})
+@SpringBootTest
```
pshapiro4broad, line 40: This test doesn't seem to need spring boot. Using a direct mock test would be simpler, and then you could remove any `verify()` below that verifies an API that was mocked (due to strict stubbing). Something like:

```
@Tag("Unit")
@ExtendWith(MockitoExtension.class)
class DrsResolutionServiceTest {

  @Mock private DrsApiFactory drsApiFactory;
  @Mock private DrsProviderService drsProviderService;
  @Mock private AuthService authService;
  @Mock private AuditLogger auditLogger;
  @Mock private Executor asyncExecutor;
  private DrsResolutionService drsResolutionService;

...

  @BeforeEach
  void before() {
    drsResolutionService =
        new DrsResolutionService(
            drsApiFactory, drsProviderService, authService, auditLogger, asyncExecutor);
    when(uriComponents.getHost()).thenReturn("host.com");
    when(uriComponents.getPath()).thenReturn(PATH);
    when(drsApiFactory.getApiFromUriComponents(eq(uriComponents), any(DrsProvider.class)))
        .thenReturn(drsApi);
  }
```

should work, and then you can remove cases like `verify(authService).fetchPassports(TOKEN);` from the tests

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -220,11 +222,10 @@ private DrsObject fetchObjectInfo(
     var sendMetadataAuth = drsProvider.isMetadataAuth();
 
     var objectId = getObjectId(uriComponents);
-    log.info(
-        "Requesting DRS metadata for {} with auth required {} from host {}",
-        drsUri,
-        sendMetadataAuth,
-        uriComponents.getHost());
+    String drsRequestLogMessage =
```
s-rubenstein, line 225: Is there a reason not to in-line this? Just stylistic?

s-rubenstein (approved):

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
okotsopoulos: Great question, I defined it for reuse below in the cases where we run into issues with passports.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
okotsopoulos: I agree with you… this seems like a similar issue to what we encountered with calling the TPS client in TDR.  I was surprised that it didn't enforce the presence of a required argument.

service/src/test/java/bio/terra/drshub/services/DrsResolutionServiceTest.java
okotsopoulos: Good shout!  I'll simplify this test in a follow-on PR.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
nmalfroy: I've been poking a little...It's because (I think) the code gen only creates models for objects in the `components.schemas` part of the openapi yaml and these are in `components.requestBodies`.  Trying to see if there's a way to have it generate models for those too

PR #83, opened 2023-11-06
[DR-3228] Specify which cloud to resolve from

samanehsan:
When a user requests an `access_url` for a DRS URI, DRSHub goes through the access methods defined in the provider configuration and returns the first match. For example, in the [configuration for TDR](https://github.com/DataBiosphere/terra-drs-hub/blob/efd40df660bc36d36f615a8aad7d033bb8cce90f/service/src/main/resources/application.yml#L53) the first access method that's listed is "gs", so if a DRS URI has an access url for both "gs" and "https" (azure) it will always return the access url for "gs".

This updates the behavior so that if a user specifies the `resolveFrom` parameter (either `azure`, `gs` or `s3`), DRSHub will return the `access_url` for that cloud if it exists, otherwise it will return the first one that it finds. So if a user specifies "gs" as the `resolveFrom` field but there is only an azure `access_url`, the azure one will get returned. If the field is not specified, the behavior of the endpoint stays the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Request bodies should look like
 ```json
 {
   "url": "string",
+  "cloudPlatform": ["gs", "azure","s3", null],
   "fields": ["string"]
 }
 ```
--- a/README.md
+++ b/README.md
@@ -71,6 +71,9 @@ timeUpdated
 googleServiceAccount
 ```
 
+The `cloudPlatform` field is optional and can be used to specify the preferred cloud platform to use for returning a signed URL.
+If no option is found for the specified cloud platform, an attempt will be made to return a signed URL from a fall-back cloud platform.
+
 ## Architecture
 DrsHub is a Java 17 Spring Boot application running in Kubernetes. As it simply resolves urls and doesn't have any state, it has no database. For developer convenience, a Swagger UI is provided.
 
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -188,12 +188,19 @@ components:
       required: [url]
       properties:
         url:
+          description: The DRS URL to resolve
           type: string
         cloudPlatform:
+          description: Which cloud platform to prefer when resolving a DRS URL (options include "gs", "azure", and "s3").
+           If an access url does not exist in that cloud platform, an alternative will be returned if one exists. If
+           this field is not specified, the first access url found will be returned.
           type: string
           nullable: true
           enum: [azure, gs, s3]
         fields:
+          description: Which information to include about the Drs Object. By default the response includes
+           bucket, contentType, fileName, gsUri, hashes, localizationPath, name, size, timeCreated, timeUpdated
+           and googleServiceAccount. The accessUrl and bondProvider will only be included if specified.
           type: array
           items:
             type: string
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -192,7 +192,7 @@ components:
         resolveFrom:
           type: string
           nullable: true
-          enum: [aws, azure, gs]
+          enum: [azure, gs, s3]
         fields:
           type: array
           items:
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -22,9 +22,9 @@ paths:
           $ref: '#/components/responses/ServerError'
   /api/v4/gcs/getSignedUrl:
     post:
-      summary: Get a signed URL for a GCS Object, signed the caller's fence service account. 
+      summary: Get a signed URL for a GCS Object, signed the caller's fence service account.
         The signed URL is active for 1 hour.
-        The bucket and object in the request body are optional, but providing them allows DRSHub to bypass the 
+        The bucket and object in the request body are optional, but providing them allows DRSHub to bypass the
         DRS object resolution and to just sign the URL using the fence account associated with the DRS Provider
         in the dataObjectUri.
       tags: [ gcs ]
@@ -189,6 +189,10 @@ components:
       properties:
         url:
           type: string
+        resolveFrom:
+          type: string
+          nullable: true
+          enum: [aws, azure, gs]
         fields:
           type: array
           items:
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -189,7 +189,7 @@ components:
       properties:
         url:
           type: string
-        resolveFrom:
+        cloudPlatform:
           type: string
           nullable: true
           enum: [azure, gs, s3]
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
+import java.util.function.Predicate;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.scheduling.annotation.Async;
@@ -307,7 +308,8 @@ public class DrsResolutionService {
         Optional.ofNullable(uriComponents.getPath()).orElse(""), StandardCharsets.UTF_8);
   }
 
-  private Optional<AccessMethod> getAccessMethod(DrsObject drsResponse, DrsProvider drsProvider, ResolveFromEnum resolveFrom) {
+  private Optional<AccessMethod> getAccessMethod(
+      DrsObject drsResponse, DrsProvider drsProvider, ResolveFromEnum resolveFrom) {
     Optional<AccessMethod> accessMethod = Optional.empty();
     if (!isEmpty(drsResponse)) {
       List<AccessMethod> accessMethods = getAccessMethods(drsResponse, drsProvider);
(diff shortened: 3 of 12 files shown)


samanehsan:
@nmalfroy brought up a good point that if there is no `access_url` for the cloud that the user wants to resolve from, DRSHub should return another one if it exists. So if a user wants to resolve from `gs` but there is only an azure `access_url`, return the azure one.

The other thing is that a user may want to specify which region they are in. If a user wants to resolve from `gs`, but there are only access urls in `s3` and `azure`, the one that is geographically closer could be returned. However, there may be other options in the future, such as an `access_url` in hot vs. cold storage. In that case we may not want to default to the closer file if it is in cold storage, and there is an alternative that is in hot storage.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -302,19 +307,51 @@ static String getObjectId(UriComponents uriComponents) {
...
-        if (matchingMethod.isPresent()) {
-          return matchingMethod;
-        }
+                .toList();
+        accessMethods.addAll(matchingMethod);
       }
     }
-    return Optional.empty();
+    return accessMethods;
   }
```
pshapiro4broad: In general if you have `List l; x.stream().stream(add to l)` you can instead write it using `flatMap()`. In this case you can write:
```
  private List<AccessMethod> getAccessMethods(DrsObject drsResponse, DrsProvider drsProvider) {
    if (isEmpty(drsResponse)) {
      return List.of();
    }
    return drsProvider.getAccessMethodConfigs().stream()
        .flatMap(methodConfig -> drsResponse.getAccessMethods().stream()
            .filter(m -> methodConfig.getType().getReturnedEquivalent() == m.getType()))
        .toList();
  }
```

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -302,19 +307,51 @@ static String getObjectId(UriComponents uriComponents) {
...
+              .filter(m -> m.getAccessId().startsWith("az"))
+              .findFirst();
+    } else {
+      accessMethod =
+          accessMethods.stream()
+              .filter(m -> m.getType().toString().equals(resolveFrom.toString()))
+              .findFirst();
+    }
+    return accessMethod;
+  }
```
pshapiro4broad: Since both `if` branches are the same except for the `filter()` expressions you could make the `if` branches create the filter and only filter once.
```
  private Optional<AccessMethod> getAccessMethodForCloud(List<AccessMethod> accessMethods, ResolveFromEnum resolveFrom) {
    Predicate<AccessMethod> filter;
    if (resolveFrom == ResolveFromEnum.AZURE) {
      filter = m -> m.getAccessId() != null && m.getAccessId().startsWith("az");
    } else {
      filter = m -> m.getType().toString().equals(resolveFrom.toString());
    }
    return accessMethods.stream().filter(filter).findFirst();
  }
```

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -302,19 +307,51 @@ static String getObjectId(UriComponents uriComponents) {
...
+  private Optional<AccessMethod> getAccessMethod(DrsObject drsResponse, DrsProvider drsProvider, ResolveFromEnum resolveFrom) {
+    Optional<AccessMethod> accessMethod = Optional.empty();
+    if (!isEmpty(drsResponse)) {
+      List<AccessMethod> accessMethods = getAccessMethods(drsResponse, drsProvider);
+      if (resolveFrom != null) {
+        accessMethod = getAccessMethodForCloud(accessMethods, resolveFrom);
+      }
+      // if there is no access method matching the resolveFrom cloud, or
+      // if the resolveFrom cloud was not specified, return a different access method
+      if (accessMethod.isEmpty() & !accessMethods.isEmpty()) {
```
pshapiro4broad: I think you meant to write `&&` here
```suggestion
      if (accessMethod.isEmpty() && !accessMethods.isEmpty()) {
```

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
samanehsan: That's a good idea!

common/openapi.yml
```diff
@@ -189,6 +189,10 @@ components:
       properties:
         url:
           type: string
+        cloudPlatform:
```
nmalfroy, line 193: From talking before: it'd be great to have descriptions on these fields and it gives a good way of explaining the behavior of `cloudPlatform`

service/src/main/java/bio/terra/drshub/util/AccessMethodUtils.java
```diff
@@ -0,0 +1,58 @@
...
+  public static Optional<AccessMethod> getAccessMethod(
+      DrsObject drsResponse, DrsProvider drsProvider, CloudPlatformEnum cloudPlatform) {
+    Optional<AccessMethod> accessMethod = Optional.empty();
+    if (!isEmpty(drsResponse)) {
+      List<AccessMethod> accessMethods = getAccessMethods(drsResponse, drsProvider);
+      if (cloudPlatform != null) {
+        accessMethod = getAccessMethodForCloud(accessMethods, cloudPlatform);
+      }
+      // if there is no access method matching the cloudPlatform, or
+      // if the cloudPlatform was not specified, return a different access method
```
nmalfroy:
```suggestion
      // if the cloudPlatform was not specified, return the first access method
```

service/src/main/java/bio/terra/drshub/util/AccessMethodUtils.java
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+    return accessMethod;
+  }
+
+  public static Optional<AccessMethod> getAccessMethodForCloud(
+      List<AccessMethod> accessMethods, CloudPlatformEnum cloudPlatform) {
+    Predicate<AccessMethod> filter;
+    if (cloudPlatform.equals(CloudPlatformEnum.AZURE)) {
+      filter = m -> m.getAccessId() != null && m.getAccessId().startsWith("az");
```
nmalfroy, line 40: We should put a comment that this is TDR behavior.  Not every DRS provider will prefix their access methods with `az`

service/src/main/java/bio/terra/drshub/util/AccessMethodUtils.java
```diff
@@ -0,0 +1,58 @@
...
+    Optional<AccessMethod> accessMethod = Optional.empty();
+    if (!isEmpty(drsResponse)) {
+      List<AccessMethod> accessMethods = getAccessMethods(drsResponse, drsProvider);
+      if (cloudPlatform != null) {
+        accessMethod = getAccessMethodForCloud(accessMethods, cloudPlatform);
+      }
+      // if there is no access method matching the cloudPlatform, or
+      // if the cloudPlatform was not specified, return a different access method
+      if (accessMethod.isEmpty() && !accessMethods.isEmpty()) {
+        accessMethod = Optional.of(accessMethods.get(0));
```
nmalfroy: To be consistent with below, I would do:
```suggestion
        accessMethod = accessMethods.stream().findFirst();
```

nmalfroy: And if you use that syntax, then you don't even need the `&& !accessMethods.isEmpty()` above.

You could actually just write the if + return part as:
```return accessMethod.orElseGet(() -> accessMethods.stream().findFirst())```

**Review by nmalfroy — APPROVED** [review]
Looks good!  Couple small comments that would be great to get in before merge

File: README.md [hunk 8]
```diff
@@ -35,6 +35,7 @@ Request bodies should look like
 ```json
 {
   "url": "string",
+  "cloudPlatform": ["gs", "azure","s3", null],
```
okotsopoulos, line 38: @nmalfroy -- can we slightly expand the README to indicate that this field is optional and what the fallback behavior is, similar to what we do for `fields` below?

service/src/main/java/bio/terra/drshub/util/AccessMethodUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import io.github.ga4gh.drs.model.AccessMethod;
+import io.github.ga4gh.drs.model.DrsObject;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Predicate;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+
+@Component
+@Slf4j
```
okotsopoulos, line 16: I don't think this annotation is necessary at this time, since nothing is logged.

service/src/main/java/bio/terra/drshub/util/AccessMethodUtils.java
```diff
@@ -0,0 +1,59 @@
...
+  public static Optional<AccessMethod> getAccessMethod(
+      DrsObject drsResponse, DrsProvider drsProvider, CloudPlatformEnum cloudPlatform) {
+    Optional<AccessMethod> accessMethod = Optional.empty();
+    if (!isEmpty(drsResponse)) {
+      List<AccessMethod> accessMethods = getAccessMethods(drsResponse, drsProvider);
+      if (cloudPlatform != null) {
+        accessMethod = getAccessMethodForCloud(accessMethods, cloudPlatform);
+      }
+      // if there is no access method matching the cloudPlatform, or
+      // if the cloudPlatform was not specified, return the first access method
```
okotsopoulos, line 27: This could be nice to break out into a top-level docstring instead.

service/src/main/java/bio/terra/drshub/util/AccessMethodUtils.java
okotsopoulos: I'm a bit uneasy about baking in provider-specific handling into code rather than config, but I'm not sure what the clean alternative is.

okotsopoulos (approved):
@nmalfroy (tag since you're shepherding this PR through), this looks good!  Nice work, Saman!

None of my comments are major, the one I feel more strongly about is updating the README a little more.

Once this is merged, I'd like to conduct a round of DRS scale testing in Terra Dev, shouldn't impact anything but good to be sure.

service/src/main/java/bio/terra/drshub/util/AccessMethodUtils.java
nmalfroy: Yeah, unfortunately, there's nothing in the DRS spec to explicitly return the cloud type...

nmalfroy:
I'm going to merge.  I've added some unit tests but it keeps squawking and I think we're in good shape

PR #84, opened 2023-11-13
[DR-3354] Fix retrieving an access url when a user has no passport

samanehsan:
**Ticket:** https://broadworkbench.atlassian.net/browse/DR-3354

If a DRS object can be accessed with a passport but the user does not have one, DRSHub will still attempt to get an access url for that object but [pass in an empty string](https://github.com/DataBiosphere/terra-drs-hub/blob/dev/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java#L289) as the passport. DRSHub catches the exception, logs an error message, and (in the case of TDR) falls back to use authentication with a bearer token. It's not necessary to call this endpoint if the user does not have a passport, so this PR adds in a check to make sure it exists.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -285,8 +285,11 @@ public class DrsResolutionService {
             }
             case PASSPORTAUTH -> {
               try {
-                yield drsApi.postAccessURL(
-                    Map.of("passports", auth.orElse(List.of(""))), objectId, accessId);
+                if (auth.isPresent()) {
+                  yield drsApi.postAccessURL(Map.of("passports", auth.get()), objectId, accessId);
+                } else {
+                  yield null;
+                }
               } catch (RestClientException e) {
                 log.error(
                     "Passport authorized request failed for {} with error {}",
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -285,11 +285,9 @@ public class DrsResolutionService {
             }
             case PASSPORTAUTH -> {
               try {
-                if (auth.isPresent()) {
-                  yield drsApi.postAccessURL(Map.of("passports", auth.get()), objectId, accessId);
-                } else {
-                  yield null;
-                }
+                yield auth.map(
+                        a -> drsApi.postAccessURL(Map.of("passports", a), objectId, accessId))
+                    .orElse(null);
               } catch (RestClientException e) {
                 log.error(
                     "Passport authorized request failed for {} with error {}",
--- a/service/src/test/java/bio/terra/drshub/controllers/DrsHubApiControllerTest.java
+++ b/service/src/test/java/bio/terra/drshub/controllers/DrsHubApiControllerTest.java
@@ -128,7 +128,7 @@ public class DrsHubApiControllerTest extends BaseTest {
     mockExternalcredsApi("ras", TEST_ACCESS_TOKEN, Optional.empty());
 
     mockBondLinkAccessTokenApi(
-        cidProviderHost.drsProvider().getBondProvider().get(),
+        cidProviderHost.drsProvider().getBondProvider().orElseThrow(),
         TEST_ACCESS_TOKEN,
         TEST_BOND_SA_TOKEN);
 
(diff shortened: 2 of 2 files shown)


service/src/test/java/bio/terra/drshub/controllers/DrsHubApiControllerTest.java
```diff
@@ -144,6 +147,8 @@ void testCallsCorrectEndpointsWhenOnlyAccessUrlRequestedWithPassportsUsingFallba
 
     // need an extra verify because nothing in the mock cares that bearer token is set or not
     verify(drsApi).setBearerToken(TEST_BOND_SA_TOKEN);
+    verify(drsApi, never())
```
okotsopoulos: I'm not sure this verification gives us full peace of mind -- could you expand it so that we're verifying that `postAccessURL` is never called with any inputs under this test set up?

Would you also mind adding a comment for this verification too?  I wish `verify` let us add a reason to the call!

okotsopoulos (approved):
Thank you for fixing this! I put a request for testing but otherwise looks good!

service/src/test/java/bio/terra/drshub/controllers/DrsHubApiControllerTest.java
okotsopoulos: PS. While you're in the neighborhood, it would be nice to add another test to handle the case where passport is a supported means of auth, but `postAccessURL` throws an exception, to make sure that we gracefully fall back in that case too.  Maybe there's a test that already does that!

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -285,8 +285,11 @@ private AccessURL fetchDrsObjectAccessUrl(
             }
             case PASSPORTAUTH -> {
               try {
-                yield drsApi.postAccessURL(
-                    Map.of("passports", auth.orElse(List.of(""))), objectId, accessId);
+                if (auth.isPresent()) {
+                  yield drsApi.postAccessURL(Map.of("passports", auth.get()), objectId, accessId);
+                } else {
+                  yield null;
+                }
```
pshapiro4broad: When possible I think it's best to avoid `isPresent()` ... `get()` to avoid duplicating the optional expression. Instead use either `map()` (when producing a value) or `ifPresent()` (when producing side effects only). In this case `map()` should work
```suggestion
                yield auth.map(
                        a -> drsApi.postAccessURL(Map.of("passports", a), objectId, accessId))
                    .orElse(null);
```

service/src/test/java/bio/terra/drshub/controllers/DrsHubApiControllerTest.java
```diff
@@ -144,6 +146,48 @@ void testCallsCorrectEndpointsWhenOnlyAccessUrlRequestedWithPassportsUsingFallba
...
+            cidProviderHost.dnsHost(), drsObject, accessId, TEST_ACCESS_URL);
+
+    when(drsApi.postAccessURL(
+            Map.of("passports", List.of(TEST_PASSPORT)), drsObject.getId(), accessId))
+        .thenThrow(new RestClientException("Failed to retrieve access url with passport"));
+
+    mockExternalcredsApi("ras", TEST_ACCESS_TOKEN, Optional.of(TEST_PASSPORT));
+
+    mockBondLinkAccessTokenApi(
+        cidProviderHost.drsProvider().getBondProvider().get(),
```
pshapiro4broad: This is just test code but using `get()` without `isPresent()` is unsafe; to make the unsafe-ness explicit I would use `orElseThrow()` instead of `get()`.
```suggestion
        cidProviderHost.drsProvider().getBondProvider().orElseThrow(),
```

service/src/test/java/bio/terra/drshub/controllers/DrsHubApiControllerTest.java
```diff
@@ -144,6 +146,48 @@ void testCallsCorrectEndpointsWhenOnlyAccessUrlRequestedWithPassportsUsingFallba
...
+            TEST_ACCESS_TOKEN,
+            cidProviderHost.compactUriPrefix(),
+            drsObject.getId(),
+            List.of(Fields.ACCESS_URL))
+        .andExpect(status().isOk())
+        .andExpect(
+            content()
+                .json(
+                    objectMapper.writeValueAsString(Map.of(Fields.ACCESS_URL, TEST_ACCESS_URL)),
+                    true));
```
pshapiro4broad: I think tests should avoid duplication like other code we need to maintain. In this case this code fragment is repeated three other times in this file, and can be refactored into a common method.

pshapiro4broad (approved):
Looks OK, a few minor comments

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
samanehsan: This looks much nicer, thanks!

PR #85, opened 2023-11-15
[DR-3361] Only log events to Bard when enabled

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3361

Possible contributor to degraded DrsHub performance following initial 8/15/23 load tests: we now log successful DRS resolutions in Bard.

To help verify the extent of Bard event logging’s impact on performance, we want to selectively disable it in DrsHub dev for additional scale tests.

Logging events to Bard is now controlled by new property `drshub.bardEventLoggingEnabled` which defaults to true, and can be overridden via environment variable `BARD_EVENT_LOGGING_ENABLED`.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
+++ b/service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
@@ -36,4 +36,8 @@ public interface DrsHubConfigInterface {
   // If this is true, then we will track calls to the Bard API in MixPanel in addition to the
   // BigQuery Data warehouse.
   Boolean trackInMixPanel();
+
+  // Only log events to Bard when enabled: this is exposed to help debug performance regressions
+  // following the introduction of Bard logging.
+  Boolean bardEventLoggingEnabled();
 }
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
```diff
@@ -36,4 +36,8 @@ public interface DrsHubConfigInterface {
   // If this is true, then we will track calls to the Bard API in MixPanel in addition to the
   // BigQuery Data warehouse.
   Boolean trackInMixPanel();
+
+  // Only log events to Bard when enabled: this is exposed to help debug performance regressions
+  // following the introduction of Bard logging.
+  Boolean bardEventLoggingEnabled();
```
pshapiro4broad, line 42: This is following the existing convention, but I don't see why this interface uses boxed types. It looks like the generated code (`DrsHubConfig`) requires that a property is set before its getter is called, so it will always have a non-null value. Something to consider for a future PR.

pshapiro4broad (approved):

s-rubenstein (approved):

tlangs (approved):

service/src/main/java/bio/terra/drshub/config/DrsHubConfigInterface.java
okotsopoulos: Yeah, good question. I wasn't sure why it was doing that either but opted to follow convention, since there's a good chance we'll ultimately remove this property support anyway.

PR #88, opened 2023-11-27
[DR-3369] ApiClients share RestTemplates

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3369

Each `DrsProvider` will still have new `ApiClients` made on request, but now they will all share the same backing `RestTemplate`. This brings behavior in line with the caching present at our last successful scale test in August '23, without sharing headers between requests (the impetus for the rollback).

### DIFF
--- a/.github/workflows/publish-branch-image.yml
+++ b/.github/workflows/publish-branch-image.yml
@@ -7,18 +7,18 @@ env:
 
 jobs:
   publish-branch-image-job:
+    outputs:
+      tag: ${{ steps.tag.outputs.tag }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


okotsopoulos:
I'm closing this PR and reopening with a new branch name: I was [unable to successfully build an image](https://github.com/DataBiosphere/terra-drs-hub/actions/runs/7037014084/job/19150871692) using the branch's existing name:

> Invalid image reference gcr.io/broad-dsp-gcr-public/terra-drs-hub:okotsopo/DR-3369-apiclients-should-share-resttemplates, perhaps you should check that the reference is formatted correctly according to https://docs.docker.com/engine/reference/commandline/tag/#extended-description
> … **For example, slash-separated name components cannot have uppercase letters**

PR #92, opened 2023-11-29
[DR-3369] ApiClients share RestTemplates

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3369

Each DrsProvider will still have new ApiClients made on request, but now they will all share the same backing RestTemplate. This brings behavior in line with the caching present at our last successful scale test in August '23, without sharing headers between requests (the impetus for the rollback).

**Performance Testing**

Note: our most recent 20k DRS URI dev-tier scale test in [Grafana](https://grafana.dsp-devops.broadinstitute.org/goto/EiPs66HIk?orgId=1) shows a peak request rate of 54 requests / second.

I added in two new performance test configurations for testing 50 and 100 concurrent TDR DRS URI resolution user journeys, respectively.  Each user journey resolves 4 DRS URIs sequentially, so these tests perform a total of 200 and 400 DRS URI resolutions, respectively.

These performance tests can be run off of a local DrsHub (I filed https://broadworkbench.atlassian.net/browse/DR-3380 in response to the discovery that DrsHub perf deployment no longer exists).

I saw modest improvements to latency when running these tests off of this feature branch versus running them off of dev, assessed via running `curl http://localhost:9098/actuator/prometheus | grep /api/v4/drs/resolve` after completion.  This is enough to feel comfortable moving forward with merging this PR but as the infrastructure and test are different enough from the workflow scale test, I am not positive what the outcome of that test will be after merge.

### DIFF
--- a/integration/src/main/resources/configs/perf/ResolveTdrDrs100.json
+++ b/integration/src/main/resources/configs/perf/ResolveTdrDrs100.json
@@ -7,8 +7,8 @@
   "testScripts": [
     {
       "name": "ResolveTdrDrs",
-      "numberOfUserJourneyThreadsToRun": 110,
-      "userJourneyThreadPoolSize": 100,
+      "numberOfUserJourneyThreadsToRun": 100,
+      "userJourneyThreadPoolSize": 110,
       "expectedTimeForEach": 10,
       "expectedTimeForEachUnit": "SECONDS"
     }
--- a/integration/src/main/resources/configs/perf/ResolveTdrDrs100.json
+++ b/integration/src/main/resources/configs/perf/ResolveTdrDrs100.json
@@ -1,0 +1,17 @@
+{
+  "name": "ResolveTdrDrs100",
+  "description": "100 concurrent attempts to resolve DRS URIs from TDR.",
+  "serverSpecificationFile": "drshub-local.json",
+  "kubernetes": {},
+  "application": {},
+  "testScripts": [
+    {
+      "name": "ResolveTdrDrs",
+      "numberOfUserJourneyThreadsToRun": 110,
+      "userJourneyThreadPoolSize": 100,
+      "expectedTimeForEach": 10,
+      "expectedTimeForEachUnit": "SECONDS"
+    }
+  ],
+  "testUserFiles": ["scarlett.json"]
+}
--- a/integration/src/main/resources/suites/FullPerf.json
+++ b/integration/src/main/resources/suites/FullPerf.json
@@ -1,9 +1,11 @@
 {
   "name": "FullPerf",
-  "description": "All perf tests",
-  "serverSpecificationFile": "drshub-perf.json",
+  "description": "All perf tests locally in the absence of a live perf environment (see: DR-3380)",
+  "serverSpecificationFile": "drshub-local.json",
   "testConfigurationFiles": [
     "perf/GetStatus.json",
-    "perf/GetVersion.json"
+    "perf/GetVersion.json",
+    "perf/ResolveTdrDrs50.json",
+    "perf/ResolveTdrDrs100.json"
   ]
 }
--- a/service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
@@ -3,18 +3,11 @@ package bio.terra.drshub.services;
 import bio.terra.drshub.config.DrsProvider;
 import bio.terra.drshub.models.DrsApi;
 import io.github.ga4gh.drs.client.ApiClient;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 import lombok.extern.slf4j.Slf4j;
-import nl.altindag.ssl.SSLFactory;
-import nl.altindag.ssl.util.Apache4SslUtils;
-import nl.altindag.ssl.util.PemUtils;
-import org.apache.http.config.Registry;
-import org.apache.http.config.RegistryBuilder;
-import org.apache.http.conn.socket.ConnectionSocketFactory;
-import org.apache.http.impl.client.CloseableHttpClient;
-import org.apache.http.impl.client.HttpClients;
-import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
-import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
 import org.springframework.stereotype.Service;
 import org.springframework.web.client.RestTemplate;
 import org.springframework.web.util.UriComponents;
@@ -23,19 +16,25 @@ import org.springframework.web.util.UriComponents;
 @Slf4j
 public class DrsApiFactory {
 
-  private static final Integer CONNECTION_POOL_SIZE = 500;
+  private final RestTemplateFactory restTemplateFactory;
+
+  /**
+   * We create a new ApiClient for each call so that headers and tokens are not shared between
+   * calls, but each DRS provider can safely share one RestTemplate among its ApiClients.
+   */
+  private final Map<String, RestTemplate> restTemplateCache =
+      Collections.synchronizedMap(new HashMap<>());
+
+  public DrsApiFactory(RestTemplateFactory restTemplateFactory) {
+    this.restTemplateFactory = restTemplateFactory;
+  }
 
   public DrsApi getApiFromUriComponents(UriComponents uriComponents, DrsProvider drsProvider) {
-    log.info(
+    log.debug(
         "Creating new DrsApi client for host '{}', for DRS Provider '{}'",
         uriComponents.getHost(),
         drsProvider.getName());
-    var mTlsConfig = drsProvider.getMTlsConfig();
-    var drsClient =
-        mTlsConfig == null
-            ? new ApiClient(makeRestTemplateWithPooling())
-            : new ApiClient(
-                makeMTlsRestTemplateWithPooling(mTlsConfig.getCertPath(), mTlsConfig.getKeyPath()));
+    var drsClient = new ApiClient(getOrCreateRestTemplate(drsProvider));
 
     drsClient.setBasePath(
         drsClient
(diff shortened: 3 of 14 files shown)


integration/src/main/resources/suites/FullPerf.json
```diff
@@ -1,9 +1,11 @@
 {
   "name": "FullPerf",
-  "description": "All perf tests",
-  "serverSpecificationFile": "drshub-perf.json",
+  "description": "All perf tests locally in the absence of a live perf environment (see: DR-3380)",
```
okotsopoulos, line 3: Filed https://broadworkbench.atlassian.net/browse/DR-3380 for complete clean-up of artifacts from now-defunct perf environment, and planning how we'd like to run these tests in the future.

service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
```diff
@@ -45,36 +44,20 @@ public DrsApi getApiFromUriComponents(UriComponents uriComponents, DrsProvider d
...
-        HttpClients.custom()
-            .setSSLSocketFactory(socketFactory)
-            .setConnectionManager(poolingConnManager)
-            .build();
-    HttpComponentsClientHttpRequestFactory factory =
-        new HttpComponentsClientHttpRequestFactory(httpClient);
-    return new RestTemplate(factory);
+  private RestTemplate getOrCreateRestTemplate(DrsProvider drsProvider) {
+    var name = drsProvider.getName();
+    if (restTemplateCache.containsKey(name)) {
```
pshapiro4broad: It'd be better to avoid work here if possible. Normally `.containsKey()` is lightweight but in a synchronized map it requires a mutex lock.
```suggestion
    if (log.isDebugEnabled() && restTemplateCache.containsKey(name)) {
```

service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
```diff
@@ -23,19 +16,25 @@
...
 public class DrsApiFactory {
 
-  private static final Integer CONNECTION_POOL_SIZE = 500;
+  private final RestTemplateFactory restTemplateFactory;
+
+  /**
+   * We create a new ApiClient for each call so that headers and tokens are not shared between
+   * calls, but each DRS provider can safely share one RestTemplate among its ApiClients.
+   */
+  private final Map<String, RestTemplate> restTemplateCache =
```
pshapiro4broad, line 25: Not sure what our conventions are, but I've seen a convention before that mutable state in a singleton component (like `@Service`) is stored in a `static final` field. This can make it clearer to the reader that the state is shared to all threads.

(of course, the lifetime of a static field object is different than that of a singleton object, e.g., spring `@Lazy` wouldn't affect its construction, so it's not always a good replacement form)

service/src/main/java/bio/terra/drshub/services/RestTemplateFactory.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestTemplate;
+
+@Service
+@Slf4j
+public class RestTemplateFactory {
+
+  private static final Integer CONNECTION_POOL_SIZE = 500;
```
pshapiro4broad: For an unboxed value, it's better to use an unboxed type. This makes it explicit to the reader (and the compiler) that the value can never contain `null`.
```suggestion
  private static final int CONNECTION_POOL_SIZE = 500;
```

nmalfroy: Should this be app config?

nmalfroy (approved):
LGTM!  Just maybe want to make the pool size configurable?

service/src/main/java/bio/terra/drshub/services/RestTemplateFactory.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.http.conn.socket.ConnectionSocketFactory;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestTemplate;
+
+@Service
+@Slf4j
```
pshapiro4broad: This class doesn't log so it doesn't need this annotation. But maybe the convention here is to always use it?

integration/src/main/resources/configs/perf/ResolveTdrDrs100.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "ResolveTdrDrs100",
+  "description": "100 concurrent attempts to resolve DRS URIs from TDR.",
+  "serverSpecificationFile": "drshub-local.json",
+  "kubernetes": {},
+  "application": {},
+  "testScripts": [
+    {
+      "name": "ResolveTdrDrs",
+      "numberOfUserJourneyThreadsToRun": 110,
```
samanehsan: Why is the number of user journey threads 110 instead of 100? (I'm wondering the same thing for the `ResolveTdrDrs50` test as well which has 60 user journey threads).

samanehsan (approved):
Looks great! I just had a clarifying question about the perf test configs ⛄

service/src/test/java/bio/terra/drshub/services/DrsApiFactoryTest.java
```diff
@@ -1,57 +1,80 @@
...
+import org.springframework.web.util.UriComponents;
 import org.springframework.web.util.UriComponentsBuilder;
 
 @Tag("Unit")
-public class DrsApiFactoryTest extends BaseTest {
+@ExtendWith(MockitoExtension.class)
+class DrsApiFactoryTest {
 
-  @Autowired DrsApiFactory drsApiFactory;
+  @Spy private RestTemplateFactory restTemplateFactory;
```
pshapiro4broad: In general I think we should avoid writing tests that use `spy()`, as it makes it harder to reason about what is being tested vs what is being mocked. In this case, why is `Spy` needed for `RestTemplateFactory`?

service/src/main/java/bio/terra/drshub/services/RestTemplateFactory.java
okotsopoulos: Good idea!  The previous implementation had it hard-coded but it would be nice to have control over it.

integration/src/main/resources/configs/perf/ResolveTdrDrs100.json
okotsopoulos: Oh, good catch!  I accidentally reversed these numbers and will fix… I wanted a pool size that was a bit bigger than the number of threads to ensure that they all ran simultaneously rather than waiting for space.

pshapiro4broad (approved):
This looks OK but it'd be better if we could test it without needing to use `Spy`. ApiClient makes this tricky but adding another layer in between the class under test and the ApiClient API would be one way to fix it. You'd end up with two factories, one which creates the ApiClient and the other which handles caching and creating the correct kind of rest template. Doing it would let you use mocked ApiClients in the test which would make it easier to assert that the correct arguments are being passed to the ApiClient constructor.

integration/src/main/resources/configs/perf/ResolveTdrDrs100.json
samanehsan: Oh that makes sense!

service/src/test/java/bio/terra/drshub/services/DrsApiFactoryTest.java
okotsopoulos: I also don't like using spies in tests.

I initially removed it in favor of a mock, but added it back in following a screech from [Codecov](https://github.com/DataBiosphere/terra-drs-hub/runs/19306586354) that the testing coverage had gone down significantly as a result: https://github.com/DataBiosphere/terra-drs-hub/pull/92/commits/4b491e82bcce88d9d3151c88c68f559a4aadaec7

I could split this out into three layers instead of two (existing is just `DrsApiFactory`), but I got stuck when trying to test `RestTemplateFactory` in its own unit test suite to maintain coverage.  I was unsure what to verify: on creation of these RestTemplates, they end up locking down most of their contents.  Any ideas?

service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
okotsopoulos: This isn't an existing convention in the repo, but I like it and will add it here.

service/src/test/java/bio/terra/drshub/services/DrsApiFactoryTest.java
okotsopoulos: Thanks for your help -- resolved in [83b86af](https://github.com/DataBiosphere/terra-drs-hub/pull/92/commits/83b86af48313d69a3ad8e25a02606245620b379d)!

service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
okotsopoulos: I changed this back: I don't understand completely why, but this change resulted in failing tests when running the full DrsApiFactoryTest suite.

okotsopoulos:
Note: CodeCov has flagged that unit test coverage has gone down in this PR.  I've reviewed the report, and the driving factor here is breaking out `RestTemplate` creation into `RestTemplateFactory` and mocking it in `DrsApiFactoryTest` rather than calling the real methods.

https://app.codecov.io/gh/DataBiosphere/terra-drs-hub/pull/92

I do not see how the existing coverage provided us with real value, and also don't think that writing `RestTemplateFactoryTest`s to make up the coverage would provide us with real value either.  `RestTemplate` creation is an opaque process and we'd be limited in our ability to verify behaviors.

So, I'd like to merge anyway despite the drop in coverage.  But please let me know if you have other preferences!  CC: @pshapiro4broad , @nmalfroy , @samanehsan

nmalfroy:
Thanks for looking.  I'm ok with going ahead and merging. I agree that sometimes the code cov ratchets aren't helpful and this is one of those cases

pshapiro4broad:
I'm also OK to override the codecov warnings for this PR.

service/src/main/java/bio/terra/drshub/services/DrsApiFactory.java
pshapiro4broad: Oh yeah, you might have to explicitly `clear()` it between tests since it's not reset like other state is. Another reason to leave it as-is I suppose.

PR #93, opened 2023-12-05
add scaffold layer to avoid spy

pshapiro4broad:
To avoid `spy()` and to provide better test support, add a layer between ApiClient and its arguments to we can see when it's called and with what arguments.

### DIFF
--- a/service/src/test/java/bio/terra/drshub/services/DrsApiFactoryTest.java
+++ b/service/src/test/java/bio/terra/drshub/services/DrsApiFactoryTest.java
@@ -1,20 +1,21 @@
 package bio.terra.drshub.services;
 
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.spy;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoInteractions;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
 
 import bio.terra.drshub.config.DrsProvider;
 import bio.terra.drshub.config.MTlsConfig;
+import io.github.ga4gh.drs.client.ApiClient;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.Spy;
+import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.web.client.RestTemplate;
 import org.springframework.web.util.UriComponents;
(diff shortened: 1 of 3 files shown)


service/src/test/java/bio/terra/drshub/services/DrsApiFactoryTest.java
```diff
@@ -38,43 +45,44 @@ private DrsProvider createDrsProvider() {
...
-    drsApiFactory.getApiFromUriComponents(
-        URI_COMPONENTS, createDrsProvider().setMTlsConfig(mTlsConfig));
+    var drsApi =
+        drsApiFactory.getApiFromUriComponents(
+            URI_COMPONENTS, createDrsProvider().setMTlsConfig(mTlsConfig));
+    assertThat(drsApi.getApiClient(), is(apiClient));
 
-    verify(restTemplateFactory).makeMTlsRestTemplateWithPooling(any(MTlsConfig.class));
-    verify(restTemplateFactory, never()).makeRestTemplateWithPooling();
+    verify(drsApiClientFactory).createClient(restTemplate);
```
pshapiro4broad, line 56: Actually I think this and the one on line 66 below are redundant due to strict stubbing.

service/src/test/java/bio/terra/drshub/services/DrsApiFactoryTest.java
```diff
@@ -38,43 +45,44 @@ private DrsProvider createDrsProvider() {
...
-
-    // Setting the token or headers on an ApiClient will have no effect on ApiClients created for
-    // the same DrsProvider
-    client1.setBearerToken("bearer-token");
-    verifyNoInteractions(separateApiClient);
+    drsApiFactory.getApiFromUriComponents(URI_COMPONENTS, drsProvider);
+    // If there was a cache miss, this would fail as the API would be called twice.
+    verify(restTemplateFactory).makeRestTemplateWithPooling();
 
-    client1.setHeader("test-header", "test-value");
```
okotsopoulos, line 77: These removed test assertions were providing real value, and were there in response to the bug where initial ApiClient caching inadvertently allowed headers to be shared between requests.  But in this new structure, this test is no longer the best place for them: we will want to test that DrsApiClientFactory creates new ApiClients each time, even for the same RestTemplate. I can take care of that in my main feature branch with this pulled in.

okotsopoulos (approved):
Thanks for noodling on this!  I do like this better, and will pull it into my branch.

service/src/test/java/bio/terra/drshub/services/DrsApiFactoryTest.java
pshapiro4broad: I think the new test provides an equivalent test to the original. Instead of testing the behavior of calling `DrsApi.setHeader()` (which should be covered by a test on `DrsApi`), it verifies that the same rest template is being used for the first two clients, and a different one is being used for the last client, by looking at how often `DrsApiFactory` calls `makeRestTemplateWithPooling()`. It doesn't verify that a new client is made each time, but you could by asserting

```
    verify(drsApiClientFactory, times(3)).createClient(any(RestTemplate.class));
```

or by mocking `createClient()` to create a new client each time instead of returning the same object each time.

PR #94, opened 2023-12-06
[DR-3386] CVE-2023-46589 Upgrade tomcat-embed-core to 9.0.83

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3386

Upgrade `tomcat-embed-core` to address https://nvd.nist.gov/vuln/detail/CVE-2023-46589 -- it was previously pinned to address a different vulnerability.

Note: we will be able to remove this pin when upgrading Spring Boot 2->3 and Terra Common Lib in https://broadworkbench.atlassian.net/browse/DR-3393

### DIFF
--- a/service/spring.gradle
+++ b/service/spring.gradle
@@ -15,10 +15,10 @@ dependencies {
 	// allows mocking final classes
 	testImplementation 'org.mockito:mockito-inline:5.2.0'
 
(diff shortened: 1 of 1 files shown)


nmalfroy (approved):
Good catch on the comment update

snf2ye (approved):

PR #95, opened 2023-12-09
QA-2193: Shift contact tests versioning to semver

ichengchang:
Ticket: [QA-2193](https://broadworkbench.atlassian.net/browse/QA-2193)

What:

Lift and shift to use `semver` tagging for contract tests. Change provider name to match helm chart name.

Why:

Sherlock now relies on `semver` (aka application version) to post deployment env to Pact Broker.

More details on why we're doing this shifting. Considering the creation of a ledger to trace contracts through PR and merge commits. It is also crucial for Sherlock to be able to interact with Pact Broker to record deployment environment for the appropriate release tag.

**Multiple Developers and Commits in a PR**

Multiple developers can collaborate on the same pull request (PR) and the PR can have multiple commits.

**Sherlock and Helm Charts for `drshub`**

Sherlock doesn't synchronize Helm charts for the drshub on PR commits
The synchronization is only called upon during merge commit to `dev` branch

**Continuous Publishing and Verification of Contracts**

You still want PRs to continuously verify consumer pacts.
The tagging mechanism is introduced to create a ledger of verified contracts in Pact Broker. For PRs, the tag is appended with the commit hash. When the live tag is pushed to GitHub during merge commit, the contracts will be published or verified using the live tag without hash.
This ledger allows developers to trace issues on both the consumer or provider side.

[QA-2193]: https://broadworkbench.atlassian.net/browse/QA-2193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/test/java/bio/terra/drshub/controllers/VerifyPactDrsHubApiController.java
+++ b/service/src/test/java/bio/terra/drshub/controllers/VerifyPactDrsHubApiController.java
@@ -56,6 +56,7 @@ import org.springframework.test.web.servlet.MockMvc;
 @PactBroker()
 class VerifyPactsDrsHubApiController {
   private static final String CONSUMER_BRANCH = System.getenv("CONSUMER_BRANCH");
+  private static final String CONSUMER_NAME = System.getenv("CONSUMER_NAME");
 
   @MockBean private DrsHubConfig drsHubConfig;
   @MockBean private BearerTokenFactory tokenFactory;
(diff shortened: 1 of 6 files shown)


aednichols (approved):

service/src/test/java/bio/terra/drshub/controllers/VerifyPactDrsHubApiController.java
```diff
@@ -49,9 +52,11 @@
 @Tag("Pact")
 @WebMvcTest
 @ContextConfiguration(classes = {DrsHubApiController.class, PublicApiController.class})
-@Provider("drshub-provider")
+@Provider("drshub")
```
okotsopoulos, line 55: Good catch here bringing the Pacticipant name in line with what's in Sherlock!

Do we need additional coordination with Cromwell to make this switch?  It is the only other Pacticipant in a relationship with `drshub`.

Do we need to re-register `drshub` into contract testing in Beehive?

okotsopoulos (approved):
Looks great -- my only questions are around the Pacticipant rename.

service/src/test/java/bio/terra/drshub/controllers/VerifyPactDrsHubApiController.java
ichengchang: `Cromwell` is all set. The `drshub` has been enabled for contract testing in Beehive, no need to re-register.

Thanks for doing this review so quickly!

PR #98, opened 2023-12-14
[ID-986] rename logback.xml

Shakespeared:
there was a typo in the filename, changed `lockback.xml` to `logback.xml`


Ghost-in-a-Jar (approved):

rushtong (approved):
👍🏽

okotsopoulos (approved):
Seems reasonable -- what issue(s) did this incorrect name produce?

Shakespeared:
> Seems reasonable -- what issue(s) did this incorrect name produce?

none that we noticed, I just happened to be poking around looking at integration tests and saw that it was named incorrectly. I think it might cause logs to get lost since it couldn't access the appender configs but I haven't verified that

Shakespeared:
> > Seems reasonable -- what issue(s) did this incorrect name produce?
> 
> none that we noticed, I just happened to be poking around looking at integration tests and saw that it was named incorrectly. I think it might cause logs to get lost since it couldn't access the appender configs but I haven't verified that

Mostly I just thought it would be easier to just change it than make a ticket and wait for someone else to do so :)

PR #100, opened 2024-01-02
DR-3365: Switch to send slack alerts to #Jade-alerts; Use slack app integration

snf2ye:
https://broadworkbench.atlassian.net/browse/DR-3365
_________________
TODO:

- [x] Add bot token to slack secrets via [ap_deployments PR](https://github.com/broadinstitute/terraform-ap-deployments/pull/1314)
- [x] Rename secret

### Using Slack GHA integration with Slack Apps
- Utilizing the [slack-backed GHA](https://github.com/slackapi/slack-github-action?tab=readme-ov-file#technique-2-slack-app) integration
- Created new slack app named "DRS Hub Notification" and added app to #jade-alerts channel
- Gave the bot associated with the app "chat:write" permissions and then added the bot's token to github action secrets
- App can be managed here: https://api.slack.com/apps/A06CD9LRN92

### The Notifications
A new re-usable workflow has been added that only send notifications to #jade-alerts channel on failure. 

image

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -8,216 +8,211 @@ on:
     branches: [ '**' ]
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v3
-      - name: Set up JDK 17
-        uses: actions/setup-java@v2
-        with:
-          java-version: '17'
-          distribution: 'temurin'
-
-      - name: Gradle cache
-        uses: actions/cache@v2
-        with:
-          path: |
-            ~/.gradle/caches
-            ~/.gradle/wrapper
-          key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
-          restore-keys: v1-${{ runner.os }}-gradle-${{ github.ref }}
-
-      - name: Cache SonarCloud packages
-        uses: actions/cache@v1
-        with:
-          path: ~/.sonar/cache
-          key: ${{ runner.os }}-sonar
-          restore-keys: ${{ runner.os }}-sonar
-
-      - name: Gradle build service
-        run: ./gradlew --build-cache :service:build -x test
-
-      - name: SonarQube scan service
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-        run: ./gradlew --build-cache :service:sonarqube
-
-      - name: Upload spotbugs results
-        uses: github/codeql-action/upload-sarif@main
-        with:
-          sarif_file: service/build/reports/spotbugs/main.sarif
-
-  jib:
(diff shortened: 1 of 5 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -174,22 +175,13 @@ jobs:
...
-      - name: Notify slack on failure
-        uses: 8398a7/action-slack@v3
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.POLICY_INT_WEBHOOK_URL }}
-        with:
-          status: failure
-          author_name: Build on dev (drs-hub)
-          fields: workflow,message
-          text: 'Build failed :sadpanda:'
+    if:  github.ref == 'refs/heads/dev'
```
pshapiro4broad: I wonder if this check could be done in `slack-notify-on-failure.yml`. Are there cases where we need to send a slack message on failure for a non-dev branch? I think the nightly perf and publish tasks are only run on the dev branch.

.github/workflows/build-and-test.yml
```diff
@@ -174,22 +175,13 @@ jobs:
...
     needs: [ build, jib, unit-tests, integration-tests ]
-    runs-on: ubuntu-latest
-
-    if: failure() && github.ref == 'refs/heads/dev'
-
-    steps:
-      - name: Notify slack on failure
-        uses: 8398a7/action-slack@v3
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.POLICY_INT_WEBHOOK_URL }}
```
pshapiro4broad, line 187: Will you delete this secret once the PR merges?

pshapiro4broad (approved):
LGTM. It looks like `slack-github-action` isn't widely used yet in terra but I assume this is now preferred over using a webhook?

snf2ye:
@pshapiro4broad  
> It looks like `slack-github-action` isn't widely used yet in terra but I assume this is now preferred over using a webhook?

`slackapi/slack-github-action` is preferred over `8398a7/action-slack` because the former is developed by a trusted third party (slack) instead of an unknown/untrusted third party (8398a7). If we switch to slack's GHA integration, we no longer have to maintain a forked repo (You'll usually see our repos referring to `broadinstitute/action-slack`, a fork of the other repo). 
Previously, slack's version hadn't been fully featured (as far as I could tell, at least), but I was able to reproduce the features this time. And, one better, the url provided in the new slack message actually routes you to the correct github action run. 

I plan to present this at an upcoming prod demo so that we can start to migrate.

.github/workflows/build-and-test.yml
snf2ye: Yes, thank you for the reminder!

.github/workflows/build-and-test.yml
snf2ye: Yeah, can do!

rjohanek (approved):
Looks good to me!

.github/workflows/build-and-test.yml
snf2ye: Secret has been deleted ✅

PR #103, opened 2024-01-16
ID-755 Switch over to bond in k8s for prod.

Ghost-in-a-Jar:
https://broadworkbench.atlassian.net/browse/ID-755

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -1,6 +1,7 @@
 deploy_env: ${DEPLOY_ENV:dev} # since there's a default, define this once
 
 drshub:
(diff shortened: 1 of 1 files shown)


service/src/main/resources/application.yml
```diff
@@ -1,6 +1,7 @@
 deploy_env: ${DEPLOY_ENV:dev} # since there's a default, define this once
 
 drshub:
+  bondUrl: https://bond.dsde-${deploy_env}.broadinstitute.org
```
Ghost-in-a-Jar, line 4: This will set it to k8s for all envs instead of just non-prod

em-may (approved):

tlangs (approved):

Shakespeared (approved):

PR #107, opened 2024-02-29
ID-1108 Exclude postgresql from terra-common-lib

tlangs:
DRSHub doesn't have a database, so it should exclude the postgresql dependency it gets from Terra Common Lib. Another example of Terra Common Lib doing too much.


Shakespeared (approved):

okotsopoulos (approved):

pshapiro4broad (approved):
Looks OK, although I think there should be a way to configure tcl so it uses `postgres` (and most of the other exclusions here) internally without exporting it. That is, configure the tcl project so that drshub doesn't need to do it.

PR #110, opened 2024-03-21
[ID-1126] Replace Bond with ECM

samanehsan:
**Ticket:** https://broadworkbench.atlassian.net/browse/ID-1126

**Background:** 
The Identiteam is in the process of deprecating Bond and moving all of its functionality into ECM, since both services link and manage external credentials for users. Now that ECM supports all of the Bond providers, this PR updates DrsHub so that it gets the provider access token and fence service account key from ECM.

**Changes:**
- Update the ECM library version
- Replace all calls to Bond with ECM calls
- Rename bond -> ecm where possible (the resolveDrs endpoint includes "bondProvider" as one of the "requestedFields" so I left that the same, but we could make a v2 endpoint with that field renamed)


Ghost-in-a-Jar (approved):
Looks good! We will definitely want to run the drs resolution jobs to test this in each environment.

service/src/main/resources/application.yml
```diff
@@ -29,7 +28,7 @@ drshub:
       name: NCI Cancer Research / Proteomics Data Commons (CRDC / PDC)
       hostRegex: '.*\.datacommons\.io'
       metadataAuth: false
-      bondProvider: dcf_fence
+      ecmFenceProvider: dcf_fence
```
tlangs, line 31: I think fence providers need to be `kebab-case` vs `snake_case` to match ECM's configs

tlangs (changes requested):

service/src/main/resources/application.yml
samanehsan: DrsHub maps these to kebab-case values: https://github.com/DataBiosphere/terra-drs-hub/blob/1ef7f66ce9b92d28c86db7bf993e6402d08d5649/service/src/main/java/bio/terra/drshub/models/ECMFenceProviderEnum.java#L4 I can change this to be consistent though!

samanehsan: Oh actually a dash is not allowed in the enum name. This is what it looks like in ECM:
```
public enum Provider {
    RAS("ras"),
    GITHUB("github"),
    FENCE("fence"),
    DCF_FENCE("dcf-fence"),
    KIDS_FIRST("kids-first"),
    ANVIL("anvil");
...
}
```

tlangs: ooooh, and that's whats sent to ECM!

tlangs (approved):

PR #112, opened 2024-06-25
[ID-1279] Log requested and actual cloud in Bard

samanehsan:
**Ticket:** https://broadworkbench.atlassian.net/browse/ID-1279

**What:**
We want to know when data is being egressed out of Azure. Because Azure does not have a requester pays feature like GCP, whoever is hosting/paying for storage will also need to pay any egress costs.  

**How:**
When the resolve endpoint gets called, if the response contains an `accessUrl` include the `resolvedCloud` in the Bard event properties if it is either azure or gcp.


service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -42,6 +42,7 @@ public class DrsResolutionService {
   private final DrsApiFactory drsApiFactory;
   private final DrsProviderService drsProviderService;
   private final AuthService authService;
+
```
tlangs: Is there a reason there's a newline here?

service/src/test/java/bio/terra/drshub/controllers/DrsHubApiControllerTest.java
```diff
@@ -120,7 +120,7 @@ void testFallbackWhenOnlyAccessUrlRequestedWithPassportsHasEmptyPassport() throw
     mockExternalCredsGetProviderAccessToken(
         Provider.fromValue(cidProviderHost.drsProvider().getEcmFenceProvider().get().getUriValue()),
         TEST_ACCESS_TOKEN,
-            TEST_FENCE_SA_TOKEN);
+        TEST_FENCE_SA_TOKEN);
```
tlangs, line 123: Where did these whitespace changes come from?

tlangs (approved):
Just some whitespace nits.

service/src/test/java/bio/terra/drshub/controllers/DrsHubApiControllerTest.java
Shakespeared: My drshub PR also created these same whitespace changes 🤔

Shakespeared (approved):

PR #113, opened 2024-06-27
ID-1207 DRSHub Client Service Metrics

tlangs:
We want to know which client is calling DRSHub! So, we'll have clients include an optional ~`serviceName` parameter~ `x-terra-service-id` header which will be logged along with the request. This is sent to Bard, and the event is stored in BigQuery. 

I've plugged the value into the existing Bard event, so we'll just have extra data now!


service/src/main/java/bio/terra/drshub/util/RequestUtils.java
```diff
@@ -0,0 +1,17 @@
+package bio.terra.drshub.util;
+
+import bio.terra.drshub.generated.model.ServiceName;
+import jakarta.servlet.http.HttpServletRequest;
+import java.util.Optional;
+
+public class RequestUtils {
+
+  public static Optional<ServiceName> serviceNameFromRequest(HttpServletRequest request) {
+    var header = Optional.ofNullable(request.getHeader("x-terra-service-id"));
```
Ghost-in-a-Jar, line 10: I think it should be `X-Terra-Service-ID`?

tlangs: headers are case-insensitive! This is tested in unit tests.

Ghost-in-a-Jar: oh it is case insensitive

Ghost-in-a-Jar (approved):

samanehsan (approved):

PR #114, opened 2024-06-28
[ID-1309] add IP address to DRSHub request

Shakespeared:
ticket: https://broadworkbench.atlassian.net/browse/ID-1309

We are planning to use TDR to restrict azure egress by IP range, so we are sending through the requester's IP address in DRSHub headers.


service/src/main/java/bio/terra/drshub/controllers/DrsHubApiController.java
```diff
@@ -33,7 +33,7 @@ public ResponseEntity<ResourceMetadata> resolveDrs(RequestObject body) {
 
     var userAgent = request.getHeader("user-agent");
     var forceAccessUrl = Objects.equals(request.getHeader("drshub-force-access-url"), "true");
-    var ip = request.getHeader("X-Forwarded-For");
```
Shakespeared: since headers are case insensitive, I standardized these to be all lowercase

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -257,13 +260,16 @@ AccessURL fetchDrsObjectAccessUrl(
       TypeEnum accessMethodType,
       List<DrsHubAuthorization> drsHubAuthorizations,
       AuditLogEvent.Builder auditLogEventBuilder,
+      String ip,
       String googleProject) {
 
     var drsApi = drsApiFactory.getApiFromUriComponents(uriComponents, drsProvider);
     var objectId = getObjectId(uriComponents);
+    drsApi.setHeader("x-forwarded-for", ip);
```
Shakespeared: I didn't limit this to TDR requests only since it's a header and not part of the response

common/openapi.yml
```diff
@@ -217,6 +218,8 @@ components:
           type: string
         dataObjectUri:
           type: string
+        ip:
+          type: string
```
Ghost-in-a-Jar: I think we actually dont want to add this here since IP will be in the header and not the request params

Ghost-in-a-Jar: maybe you can specify a required header in an openapi spec? I dont think i have before

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
samanehsan: Do you know what happens if the `ip` is null? Maybe we should only set it as a header if it's defined

common/openapi.yml
Shakespeared: oh yeah duh, brain fart on my part

service/src/test/java/bio/terra/drshub/util/SignedUrlTestUtils.java
```diff
@@ -51,6 +51,7 @@ public class SignedUrlTestUtils {
   public static void setupSignedUrlMocks(
       AuthService authService,
       GoogleStorageService googleStorageService,
+      String ip,
```
samanehsan: Since the `ip` isn't being used here could we remove it?

Shakespeared: it is marked as unused here but if I had to add it to get unit tests to pass

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
Shakespeared: yeah that's a good point, I'll add a check for that

samanehsan (approved):

tlangs (approved):

PR #117, opened 2024-08-07
[ID-1360] Add metadata to bard events

samanehsan:


samanehsan:
Going with a different approach (will link PR when it's up)

PR #119, opened 2024-08-08
[ID-1363] Add transaction id to resolveDrs requests

samanehsan:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1363

For every request to the resolveDrs endpoint, drshub sends a request to the drs provider (such as TDR) to get object info and optionally another request to get the access url. We need a way to join the Drshub and TDR requests in Bard so that we can then gather drs metrics using the billing profile/dataset/snapshot info.

Changes:
- Copied the `UserLoggingMetrics` class from TDR to allow additional Bard properties to be specified outside of the `TrackingInterceptor` class
- Log a transaction id (a random UUID) to Bard for resolveDrs requests
- Set this transaction id as a header on the requests to TDR (when getting the object metadata and access url)

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -28,6 +28,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -370,5 +371,9 @@ public class DrsResolutionService {
         .drsMetadata(drsMetadata)
         .drsProvider(drsProvider)
         .build();
+  }
+
+  public String getTransactionId() {
+    return UUID.randomUUID().toString();
   }
 }
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -47,7 +47,8 @@ public class DrsResolutionService {
   private final AuthService authService;
   private final AuditLogger auditLogger;
   private final UserLoggingMetrics userLoggingMetrics;
-  private static final String TRANSACTION_ID_FIELD_NAME = "transactionId";
+  public static final String TRANSACTION_ID_FIELD_NAME = "transactionId";
+  public static final String TRANSACTION_ID_HEADER_NAME = "X-Transaction-Id";
 
   @Autowired
   public DrsResolutionService(
@@ -249,7 +250,7 @@ public class DrsResolutionService {
     log.info(drsRequestLogMessage);
 
     var drsApi = drsApiFactory.getApiFromUriComponents(uriComponents, drsProvider);
-    drsApi.setHeader("x-transaction-id", transactionId);
+    drsApi.setHeader(TRANSACTION_ID_HEADER_NAME, transactionId);
     if (sendMetadataAuth) {
       // Currently, no provider needs a fence_token for metadata auth.
       // If that changes, this will need to get updated.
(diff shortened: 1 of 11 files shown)


service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -235,6 +249,7 @@ DrsObject fetchObjectInfo(
     log.info(drsRequestLogMessage);
 
     var drsApi = drsApiFactory.getApiFromUriComponents(uriComponents, drsProvider);
+    drsApi.setHeader("x-transaction-id", transactionId);
```
tlangs: `X-Transaction-ID` could probably be extracted into a variable since its referenced multiple times.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -44,17 +46,21 @@ public class DrsResolutionService {
...
       DrsProviderService drsProviderService,
       AuthService authService,
-      AuditLogger auditLogger) {
+      AuditLogger auditLogger,
+      UserLoggingMetrics userLoggingMetrics) {
     this.drsApiFactory = drsApiFactory;
     this.drsProviderService = drsProviderService;
     this.authService = authService;
     this.auditLogger = auditLogger;
+    this.userLoggingMetrics = userLoggingMetrics;
```
tlangs: DRSHub already logs its DRS Resolutions via the Audit Logger. I think we should include the transaction ID in that message, rather than publishing a separate message.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
tlangs: I don't know if we want to add the transaction id header to every request. Other DRS Providers might have similar header restrictions to us, and not allow it through. We should probably only add it if we know the request is going to TDR.

I think I might have said the opposite when we talked about it face to face, but I'm realizing now that if we're restricting what headers are allowed in requests to our systems, others are probably doing it too.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -266,7 +281,8 @@ AccessURL fetchDrsObjectAccessUrl(
       List<DrsHubAuthorization> drsHubAuthorizations,
       AuditLogEvent.Builder auditLogEventBuilder,
       String ip,
-      String googleProject) {
+      String googleProject,
+      String transactionId) {
 
     var drsApi = drsApiFactory.getApiFromUriComponents(uriComponents, drsProvider);
```
tlangs, line 284: You could probably support setting the header conditionally if its TDR from within `getApiFromUriComponents`

tlangs (changes requested):

tlangs:
Also, its convention to Capital-Kebab-Case header names. Libraries usually handle them in a case-insensitive way.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
samanehsan: This allows us to include the transaction id in the properties for existing Bard events (like how TDR does it). It's not creating an additional log message. I can include it in the audit log message as well though!

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
samanehsan: We're not checking the provider for other headers btw 😅 At least we always pass the IP address and x-user-project for access url requests.

tlangs: Oh, maybe it's fine then!

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
tlangs: Oh, do we need the `UserLoggingMetrics` so that we can inject more attributes into the Bard request?

samanehsan: Yes! Maybe there's a clearer name for this?

tlangs (approved):

Ghost-in-a-Jar (approved):

Ghost-in-a-Jar (approved):

PR #121, opened 2024-08-19
[ID-1365] Log fileName to Bard (if in response)

samanehsan:
**Ticket:** https://broadworkbench.atlassian.net/browse/ID-1365

**Purpose:**
Collaborators want metrics around usage/access of the data they store in TDR and Terra, such as which files get accessed the most.

**Changes:**
If the `fileName` field is present in the response when resolving a DRS url, include it in the properties logged to Bard.

### DIFF
--- a/service/src/test/java/bio/terra/drshub/tracking/TrackingInterceptorTest.java
+++ b/service/src/test/java/bio/terra/drshub/tracking/TrackingInterceptorTest.java
@@ -194,18 +194,24 @@ class TrackingInterceptorTest {
   }
 
   @Test
-  void testHappyPathEmittingToBardWithTransactionId() throws Exception {
+  void testHappyPathEmittingToBardWithFileName() throws Exception {
     mockBardEmissionsEnabled();
-    String transactionId = UUID.randomUUID().toString();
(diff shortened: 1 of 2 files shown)


service/src/test/java/bio/terra/drshub/tracking/TrackingInterceptorTest.java
```diff
@@ -194,18 +194,24 @@ void testHappyPathEmittingToBardWithResolvedCloudAzure() throws Exception {
   }
 
   @Test
-  void testHappyPathEmittingToBardWithTransactionId() throws Exception {
+  void testHappyPathEmittingToBardWithFileName() throws Exception {
```
samanehsan, line 197: I removed the `testHappyPathEmittingToBardWithTransactionId` because I realized I already mocked the transactionId in the `setUp` and the `userLoggingMetrics` is autowired so I do not need to set the transactionId property in the test. So after removing those unnecessary lines it looks exactly like `testHappyPathEmittingToBard`

I added `testHappyPathEmittingToBardWithFileName` too so the diff looks a bit confusing here.

Ghost-in-a-Jar (approved):

Shakespeared (approved):

PR #125, opened 2024-11-20
CORE-163 end support of googleServiceAccount field

dvoet:
https://broadworkbench.atlassian.net/browse/CORE-163

The problem in the ticket stems from ECM not properly purging expired fence service account keys. And because they never get removed, ECM never gets a new one. However, ECM never returns the expired key but returns a 404 instead. This triggers a 404 in DrsHub when the googleServiceAccount field is requested which causes failures in rawls when it is validating DRS URIs and workflows when it tries to resolve.

According to https://broadworkbench.atlassian.net/browse/PREQ-28, we no longer need these keys even though they are requested. Both rawls and cromwell should be able to handle their absence because there are already cases where they don't exist (e.g. kids first which is AWS).

I think it is better to discontinue use than to fix ECM's expiration problem.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -149,11 +149,6 @@ public class DrsResolutionService {
     var accessMethod = AccessMethodUtils.getAccessMethod(drsResponse, drsProvider, cloudPlatform);
     var accessMethodType = accessMethod.map(AccessMethod::getType).orElse(null);
 
-    if (drsProvider.shouldFetchUserServiceAccount(accessMethodType, requestedFields)) {
-      var saKey = authService.fetchUserServiceAccount(drsProvider, bearerToken);
-      drsMetadataBuilder.bondSaKey(saKey);
-    }
-
     if (drsResponse != null) {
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -149,11 +149,6 @@ private DrsMetadata fetchObject(
     var accessMethod = AccessMethodUtils.getAccessMethod(drsResponse, drsProvider, cloudPlatform);
     var accessMethodType = accessMethod.map(AccessMethod::getType).orElse(null);
 
-    if (drsProvider.shouldFetchUserServiceAccount(accessMethodType, requestedFields)) {
```
snf2ye, line 152: Can the shouldFetchUserServiceAccount method also be removed from DrsProviderInterface?

snf2ye (approved):

samanehsan (approved):

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
dvoet: eventually, yes. There is a bunch more to remove too.

PR #126, opened 2025-01-09
[DT-1122] Apply `zizmor` suggestions

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1122

### Summary

- Do not persist credentials if not needed
- Use shell variable expansion instead of [template expansion](https://woodruffw.github.io/zizmor/audits/#template-injection)

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -46,6 +46,7 @@ jobs:
         with:
           ref: ${{ inputs.ref }}
           token: ${{ secrets.BROADBOT_TOKEN }} # this allows the push to succeed later
+          persist-credentials: false
       - name: Bump the tag to a new version
         # https://github.com/DataBiosphere/github-actions/tree/master/actions/bumper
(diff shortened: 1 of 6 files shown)


rushtong (approved):
👍🏽

.github/workflows/tag.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         with:
           ref: ${{ inputs.ref }}
           token: ${{ secrets.BROADBOT_TOKEN }} # this allows the push to succeed later
+          persist-credentials: false
```
rjohanek: don't you need the credentials to persist for this action (checkout) because it's used later (bump), like did this same change cause a problem in a different PR?

rjohanek (approved):

.github/workflows/tag.yml
fboulnois: I think you're right, any place we're using the bumper will cause failures.

fboulnois: Although I'm not sure if it's specific to the data repo bumper

PR #127, opened 2025-01-14
CORE-246 Upgrading to cache v4 

kevinpalis:
The cache backend service has been rewritten from the ground up for improved performance and reliability. [actions/cache](https://github.com/actions/cache) now integrates with the new cache service (v2) APIs.

The new service will gradually roll out as of February 1st, 2025. The legacy service will also be sunset on the same date. Changes in these release are fully backward compatible.

More info here: https://github.com/marketplace/actions/cache

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -23,7 +23,7 @@ jobs:
           distribution: 'temurin'
 
       - name: Gradle cache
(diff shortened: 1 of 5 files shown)


pshapiro4broad (approved):

rushtong (approved):
👍🏽

PR #128, opened 2025-03-18
DT-1405: Update TCL and a few other dependencies

pshapiro4broad:
Various dependency / plugin updates

- update terra-common-lib to latest version
- add `CODEOWNERS`
- update gradle plugins that didn't require code changes
- update libraries which were flagged in the IDE as security risks
- update `render_config.sh` to work with gsm
- reformat using latest spotless

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,7 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+# Documentation with syntax examples:
+# https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+
+# These owners will be the default owners for everything in the repo.
+*       @DataBiosphere/dsp-data-team-eng
--- a/render_config.sh
+++ b/render_config.sh
@@ -1,26 +1,22 @@
 ENV=${1:-dev}
-VAULT_TOKEN=${2:-$(cat "$HOME"/.vault-token)}
 
-VAULT_ADDR="https://clotho.broadinstitute.org:8200"
-DRSHUB_VAULT_PATH="secret/dsde/terra/kernel/$ENV/$ENV/drshub"
-COMMON_VAULT_PATH="secret/dsde/terra/kernel/$ENV/common"
-
-VAULT_COMMAND="vault read"
+if [[ $ENV == dev ]]; then
+    GCLOUD_PROJECT=broad-dsde-dev
+elif [[ $ENV == integration ]]; then
+    GCLOUD_PROJECT=broad-dsde-qa
+else
+    echo "Invalid environment: $ENV"
+    exit 1
+fi
 
 SERVICE_OUTPUT_LOCATION="$(dirname "$0")/service/src/main/resources/rendered"
 INTEGRATION_OUTPUT_LOCATION="$(dirname "$0")/integration/src/main/resources/rendered"
 
-if ! [ -x "$(command -v vault)" ]; then
-  VAULT_COMMAND="docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN -e VAULT_ADDR=$VAULT_ADDR vault:1.7.3 $VAULT_COMMAND"
-fi
+GET_SECRET="gcloud secrets versions access latest --project $GCLOUD_PROJECT --secret"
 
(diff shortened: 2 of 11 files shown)


.github/CODEOWNERS
fboulnois, line 1: 👍

render_config.sh
```diff
@@ -1,26 +1,22 @@
...
+    exit 1
+fi
 
 SERVICE_OUTPUT_LOCATION="$(dirname "$0")/service/src/main/resources/rendered"
 INTEGRATION_OUTPUT_LOCATION="$(dirname "$0")/integration/src/main/resources/rendered"
 
-if ! [ -x "$(command -v vault)" ]; then
-  VAULT_COMMAND="docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN -e VAULT_ADDR=$VAULT_ADDR vault:1.7.3 $VAULT_COMMAND"
-fi
+GET_SECRET="gcloud secrets versions access latest --project $GCLOUD_PROJECT --secret"
```
fboulnois, line 15: It's unusual to see `$VARIABLE command` have $VARIABLE itself be the thing executing the command. Perhaps put this in a function?

```
get_secret() {
  gcloud secrets versions access latest --project "$GCLOUD_PROJECT" --secret "$1"
}

get_secret "drshub-ras-mtls-client-cert"
```

service/analysis.gradle
```diff
@@ -21,15 +21,15 @@ spotbugs {
...
+tasks.spotbugsMain {
+	reports.create("sarif") {
+		required = true
 	}
 }
 
 spotbugsTest { enabled = false }
 
-task runMinnieKenny(type: Exec) {
+tasks.register('runMinnieKenny', Exec) {
```
fboulnois, line 32: [perhaps in a separate ticket] We should remove anything related to `minnieKenny`, the secret checker is deployed by BITS to devs on their macbooks.

service/src/main/java/bio/terra/drshub/services/AuthService.java
fboulnois, line 1: This is stylistic reformatting.

fboulnois (approved):
Looks good, one suggestion:

render_config.sh
```diff
@@ -1,26 +1,22 @@
...
 
-$VAULT_COMMAND -field=value "$DRSHUB_VAULT_PATH/ras-mtls-client.crt" >"$SERVICE_OUTPUT_LOCATION/ras-mtls-client.crt"
-$VAULT_COMMAND -field=value "$DRSHUB_VAULT_PATH/ras-mtls-client.key" >"$SERVICE_OUTPUT_LOCATION/ras-mtls-client.key"
-$VAULT_COMMAND -field=swagger-client-id "$DRSHUB_VAULT_PATH/swagger-client-id" >"$SERVICE_OUTPUT_LOCATION/swagger-client-id"
+$GET_SECRET drshub-ras-mtls-client-cert >"$SERVICE_OUTPUT_LOCATION/ras-mtls-client.crt"
+$GET_SECRET drshub-ras-mtls-client-key >"$SERVICE_OUTPUT_LOCATION/ras-mtls-client.key"
+$GET_SECRET drshub-swagger-client-id >"$SERVICE_OUTPUT_LOCATION/swagger-client-id"
```
rushtong, line 19: Wow ... vault was still hanging around

rushtong (approved):
👍🏽

service/src/main/java/bio/terra/drshub/services/AuthService.java
pshapiro4broad: Yes, due to the spotless update

service/analysis.gradle
pshapiro4broad: Yes, and spotbugs should be removed too since we use sonar. I can make a ticket

render_config.sh
pshapiro4broad: I was following the existing code pattern here, and `shellcheck` doesn't flag any issues. But I can change it if you like. I read this as `<shell expansion> <stuff after expansion>` not `<command> <argument>` so it seems fine to me.

render_config.sh
pshapiro4broad: Not really. There are many scripts and READMEs that still believe it's around but this stopped working a while ago.

PR #135, opened 2025-03-19
DT-1425: Enable dependabot builds

pshapiro4broad:
To enable dependabot builds, use the `BROADBOT_TOKEN` secret instead of `GITHUB_TOKEN`, since `GITHUB_TOKEN` is not populated by https://github.com/broadinstitute/terraform-ap-deployments/blob/master/github/tfvars/databiosphere-terra-drs-hub.tfvars.

- update versions of actions to current versions.
- remove labels from dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,8 +24,5 @@ updates:
       time: "06:00"
       timezone: "America/New_York"
     target-branch: "dev"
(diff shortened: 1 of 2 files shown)


snf2ye (approved):

rushtong (approved):
👍🏽

PR #140, opened 2025-03-21
DT-1405: update to latest tcl to fix remaining otel configuration issue

pshapiro4broad:
See https://github.com/DataBiosphere/terra-common-lib/pull/229 and the other PRs that refer to it such as https://github.com/DataBiosphere/terra-workspace-manager/pull/1842

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -86,15 +86,11 @@ otel:
   sdk:
     disabled: false # set to true to disable all open telemetry features
 
-  springboot:
-    resource:
-      attributes:
-        service:
-          name: ${spring.application.name}
-
-  instrumentation:
-    logback-appender:
-      enabled: false
+  resource:
+    attributes:
+      service:
+        name: ${spring.application.name}
+        version: ${drshub.version.gitTag:unknown}
 
 server:
   max-http-request-header-size: 32KB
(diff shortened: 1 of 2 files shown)


service/src/main/resources/application.yml
```diff
@@ -86,15 +86,11 @@ otel:
...
-  springboot:
-    resource:
-      attributes:
-        service:
-          name: ${spring.application.name}
-
-  instrumentation:
-    logback-appender:
-      enabled: false
+  resource:
```
pshapiro4broad, line 89: This change is copied from https://github.com/DataBiosphere/terra-billing-profile-manager/pull/535

service/src/main/resources/application.yml
```diff
@@ -86,15 +86,11 @@ otel:
...
 
-  springboot:
-    resource:
-      attributes:
-        service:
-          name: ${spring.application.name}
-
-  instrumentation:
-    logback-appender:
-      enabled: false
```
pshapiro4broad, line 97: I believe this is no longer necessary, with the latest tcl changes in place.

snf2ye: Was this also copied from another PR?

pshapiro4broad: Not copied, but this was the original fix which disabled the appender, before we had a fix in tcl.

rjohanek (approved):
looks good to me thanks for the comments

snf2ye (approved):

PR #142, opened 2025-03-28
[CORE-413] Configure Sage DRS Provider

kevinmarete:
**Ticket:** https://broadworkbench.atlassian.net/browse/CORE-413

This PR updates `application.yml` to configure Sage's DRS provider, enabling DRShub to resolve and access Sage DRS URIs.

- **Dev:** `repo-dev.dev.sagebase.org` 
- **Prod:** `repo-prod.prod.sagebase.org`

### DIFF
--- a/service/src/main/java/bio/terra/drshub/models/ECMFenceProviderEnum.java
+++ b/service/src/main/java/bio/terra/drshub/models/ECMFenceProviderEnum.java
@@ -4,7 +4,8 @@ public enum ECMFenceProviderEnum {
   dcf_fence("dcf-fence"),
   fence("fence"),
   anvil("anvil"),
-  kids_first("kids-first");
+  kids_first("kids-first"),
+  sage("sage");
 
   private String uriValue;
 
--- a/service/src/main/java/bio/terra/drshub/models/ECMFenceProviderEnum.java
+++ b/service/src/main/java/bio/terra/drshub/models/ECMFenceProviderEnum.java
@@ -5,7 +5,7 @@ public enum ECMFenceProviderEnum {
   fence("fence"),
   anvil("anvil"),
   kids_first("kids-first"),
-  SAGE("sage");
+  sage("sage");
 
   private String uriValue;
 
--- a/service/src/main/java/bio/terra/drshub/models/ECMFenceProviderEnum.java
+++ b/service/src/main/java/bio/terra/drshub/models/ECMFenceProviderEnum.java
@@ -5,7 +5,7 @@ public enum ECMFenceProviderEnum {
   fence("fence"),
   anvil("anvil"),
   kids_first("kids-first"),
-  sage("sage");
+  SAGE("sage");
 
   private String uriValue;
 
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -57,6 +57,15 @@ drshub:
         - type: https
           auth: current_request
           fetchAccessUrl: true # Used for Azure
+    sage:
+      name: Sage Bionetworks
+      hostRegex: 'repo-([^.]+)\.\1\.sagebase\.org' # Dev: 'repo-dev.dev.sagebase.org' & Prod: 'repo-prod.prod.sagebase.org'
+      metadataAuth: false
+      ecmFenceProvider: sage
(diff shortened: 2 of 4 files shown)


service/src/main/java/bio/terra/drshub/models/ECMFenceProviderEnum.java
```diff
@@ -4,7 +4,8 @@ public enum ECMFenceProviderEnum {
   dcf_fence("dcf-fence"),
   fence("fence"),
   anvil("anvil"),
-  kids_first("kids-first");
+  kids_first("kids-first"),
+  SAGE("sage");
```
dvoet: all the others are lowercase so I would make sage lowercase too for consistency

service/src/main/resources/application.yml
```diff
@@ -160,12 +169,22 @@ drshub:
...
+        - type: s3
+          auth: fence_token
+          fetchAccessUrl: true
   signedUrlDuration: 1h
   compactIdHosts:
     "dg.712c": staging.gen3.biodatacatalyst.nhlbi.nih.gov # biodatacatalyst staging
     "dg.4dfc": nci-crdc-staging.datacommons.io # cancer research data commons
     "dg.f82a1a": gen3staging.kidsfirstdrc.org # kidsfirst
     "dg.test0": ctds-test-env.planx-pla.net # passport test
+    "dg.sage": repo-dev.dev.sagebase.org # sage bionetworks
```
dvoet: I don't think they are using a compact id so remove this and the one for prod

service/src/main/resources/application.yml
```diff
@@ -57,6 +57,15 @@ drshub:
         - type: https
           auth: current_request
           fetchAccessUrl: true # Used for Azure
+    sage:
+      name: Sage Bionetworks
+      hostRegex: 'repo-([^.]+)\.\1\.sagebase\.org' # Dev: 'repo-dev.dev.sagebase.org' & Prod: 'repo-prod.prod.sagebase.org'
```
dvoet: no need to be so fancy
```suggestion
      hostRegex: 'repo-dev\.dev\.sagebase\.org|repo-prod\.prod\.sagebase\.org'
```

service/src/main/resources/application.yml
```diff
@@ -57,6 +57,15 @@ drshub:
...
+      metadataAuth: false
+      ecmFenceProvider: SAGE
+      accessMethodConfigs:
+        - type: s3
```
dvoet, line 66: do we know that sage data is stored in s3?

service/src/main/resources/application.yml
```diff
@@ -57,6 +57,15 @@ drshub:
...
+          auth: fence_token
```
dvoet, line 67: This is confusing because this has sage does not use fence. This should probably be changed to `provider_token` and all references to `fence` be updated to be `provider`. But that is probably out of scope for this ticket.

service/src/main/resources/application.yml
```diff
@@ -160,12 +169,22 @@ drshub:
...
         certPath: rendered/ras-mtls-client.crt
+    sage:
+      name: Sage Bionetworks
+      hostRegex: 'repo-dev.dev.sagebase.org'
+      metadataAuth: false
+      ecmFenceProvider: SAGE
+      accessMethodConfigs:
+        - type: s3
+          auth: fence_token
+          fetchAccessUrl: true
```
dvoet: I don't think you need this section as it is covered by the section on line 60

dvoet (changes requested):

service/src/main/resources/application.yml
dvoet: I confirmed s3 is correct

fboulnois (commented):
I agree with @dvoet 's comments.

service/src/main/resources/application.yml
kevinmarete: Is this for all the providers or just sage?

fboulnois (approved):
👍

dvoet (approved):

s-rubenstein:
I will file a ticket about moving from `fence` to `provider`

PR #147, opened 2025-04-22
CORE-447 retry drs api calls

dvoet:
https://broadworkbench.atlassian.net/browse/CORE-447

The underlying problem is that when using passport authentication and ECM returns a 502 to TDR, TDR propagates the 502 to DrsHub, then DrsHub treats that a "passport doesn't work try something else" which returns a 403 which is the symptom the caller sees. The scale test shows just a few of these so we think a retry strategy should fix the problem instead of scaling up the infrastructure which costs $$$. Retrying is probably a good idea anyway.

I looked into using Spring's retry features but it was too much spring magic to make `DrsApi` a bean and other approaches take too much refactoring for not much benefit.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/models/DrsApi.java
+++ b/service/src/main/java/bio/terra/drshub/models/DrsApi.java
@@ -1,5 +1,6 @@
 package bio.terra.drshub.models;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.github.ga4gh.drs.api.ObjectsApi;
 import io.github.ga4gh.drs.client.ApiClient;
 import io.github.ga4gh.drs.client.auth.OAuth;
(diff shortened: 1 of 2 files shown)


service/src/test/java/bio/terra/drshub/models/DrsApiTest.java
```diff
@@ -46,4 +53,52 @@ void testSetHeader() {
 
     verify(apiClient).addDefaultHeader(name, value);
   }
+
+  @ParameterizedTest
+  @ValueSource(
+      ints = {
+        500, 502, 503, 504,
+      }) // HttpStatus.INTERNAL_SERVER_ERROR, BAD_GATEWAY, SERVICE_UNAVAILABLE, GATEWAY_TIMEOUT
```
pshapiro4broad, line 61: I think you can use java constants in this context
```suggestion
        HttpStatus.INTERNAL_SERVER_ERROR,
        HttpStatus.BAD_GATEWAY,
        HttpStatus.SERVICE_UNAVAILABLE,
        HttpStatus.GATEWAY_TIMEOUT
      })
```

rjohanek (approved):
adding retries looks good to me

service/src/main/java/bio/terra/drshub/models/DrsApi.java
```diff
@@ -17,4 +30,54 @@ public void setBearerToken(String bearerToken) {
   public void setHeader(String name, String value) {
     this.getApiClient().addDefaultHeader(name, value);
   }
+
+  @Override
+  public DrsObject postObject(Object body, String objectId) throws RestClientException {
```
pshapiro4broad, line 42: The main downside with this approach is that new APIs won't have this retry and we won't get a compile error when they're added. Changing this to embed `ObjectsApi` instead of subclassing is one way to fix this.

pshapiro4broad (approved):
This looks OK but I would prefer to see an embedded object approach for `DrsApi`, to avoid a maintainer needing to know about this feature when/if new DRS APIs are added.

Also, this adds retries for all DRS API calls, not just TDR. We are hoping/assuming that retries on these errors would be good for other DRS providers too.

service/src/test/java/bio/terra/drshub/models/DrsApiTest.java
dvoet: unfortunately no because `ValueSource` only supports primitives and those constants are not

PR #150, opened 2025-04-29
CORE-485, DT-1468 metadata auth and rename fence

dvoet:
https://broadworkbench.atlassian.net/browse/CORE-485
https://broadworkbench.atlassian.net/browse/DT-1468

The primary problem is that we are not sending an access token to the Sage metadata api call and DrsHub did not account for that case. The code change for CORE-458 is to allow configuration for different ways to authenticate for the metadata api call:
- no auth (Gen3 cases)
- current request's token (TDR)
- provider token from ECM (Sage)
- passport (as yet unused)

Note that, when implemented by the DRS provider, the OPTIONS object api governs whether or not to use a token or a passport and using a token becomes the fallback.

The secondary problem is that the use of the term `fence` was driving me nuts and causing confusion. Sorry for bundling the 2 concerns. Most of the 1 line changes are name changes.


service/src/main/java/bio/terra/drshub/config/DrsProviderInterface.java
```diff
@@ -58,37 +58,6 @@ default List<AccessMethodConfigTypeEnum> getAccessMethodConfigTypes() {
     return getAccessMethodConfigs().stream().map(ProviderAccessMethodConfig::getType).toList();
   }
 
-  /**
```
dvoet, line 61: This block and the next were unused in production code. A bunch of tests went away with this too.

service/src/main/java/bio/terra/drshub/services/AuthService.java
```diff
@@ -334,6 +330,31 @@ public String getSignedUrlForBlob(
   public void clearCaches() {
     passportCache.clear();
     serviceAccountKeyCache.clear();
-    fenceAccessTokenCache.clear();
+    providerAccessTokenCache.clear();
+  }
+
+  public String getMetadataAuthBearerToken(
```
dvoet, line 336: new function for CORE-485

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -226,7 +227,7 @@ DrsObject fetchObjectInfo(
       BearerToken bearerToken,
       List<DrsHubAuthorization> authorizations,
       String transactionId) {
-    var sendMetadataAuth = drsProvider.isMetadataAuth();
+    var sendMetadataAuth = drsProvider.metadataAuthTypeIsSet();
```
dvoet, line 230: the changes in this function are for CORE-485

snf2ye (approved):

service/src/main/resources/application.yml
```diff
@@ -9,46 +9,42 @@ drshub:
     anvil:
       name: NHGRI Analysis Visualization and Informatics Lab-space (The AnVIL)
       hostRegex: '.*\.theanvil\.io'
-      metadataAuth: false
```
rjohanek, line 12: why are the metadataAuth values not needed anymore? / what does the move from isMetaDataAuth to metadataAuthTypeIsSet do?

dvoet: I replaced `metadataAuth` (boolean) with `metadataAuthType` (enum) to figure out which token we need to send. The absence of `metadataAuthType` is interpreted as no auth is required.

rjohanek (approved):

PR #159, opened 2025-05-29
DT-1431: Update to current spring-boot

pshapiro4broad:
http://broadworkbench.atlassian.net/browse/DT-1431

This change updates drshub's spring-boot dependency to the version currently used by terra services. There is a more recent version, but that requires open telemetry changes, and other terra services haven't updated to this version yet.

The original plan was to convert drshub to use openapi codegen instead of swagger codegen, but that looks like a large task. For an immediate fix to the issue with Spring's removal of `Base64Utils`, this change provides a custom template for `HttpBasicAuth` which replaces `Base64Utils` with the JDK 8 API `Base64.getEncoder()`.

- Updated `ApiClient.mustache` to fix another spring deprecation. This required some duplication as there are two places where the drshub generates client APIs.
- Removed unused template files that were required before. It seems like the bug that required their presence has been fixed.


rushtong (approved):

service/generators.gradle
```diff
@@ -29,17 +29,17 @@ swaggerSources {
 
 	ga4gh {
 		inputFile = file('src/main/resources/vendor/data-repository-service-1.2.0.yaml')
-		code(makeClientCodeClosure('io.github.ga4gh.drs', true))
+		code(makeClientCodeClosure('io.github.ga4gh.drs'))
 	}
 
 	bard {
 		inputFile = file('src/main/resources/vendor/bard.yaml')
-		code(makeClientCodeClosure('bio.terra.bard', false))
```
pshapiro4broad, line 37: This is the one change that I'm less certain about. Previously only the drs API would use the custom template files, and now they're used for all client API codegen. I don't think this will have an effect on how these APIs are used, but I'm not sure if this usage is well exercised by the integration tests.

If we wanted to be more careful about this, I could create two different set of templates, one that fixes the deprecation issues, and another that fixes the issues and also implements the ApiClient changes for the drs client.

s-rubenstein (approved):

snf2ye:
Follow on work: https://broadworkbench.atlassian.net/browse/DT-1780

PR #226, opened 2026-03-05
[DT-3008] Add Logging for `x-user-project` header

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3008

### Summary

This PR adds logging for `x-user-project` header in `resolveDrs` and `getSignedUrl` endpoints

### DIFF
--- a/service/src/main/java/bio/terra/drshub/controllers/GcsApiController.java
+++ b/service/src/main/java/bio/terra/drshub/controllers/GcsApiController.java
@@ -33,7 +33,7 @@ public class GcsApiController implements GcsApi {
   public ResponseEntity<String> getSignedUrl(GetSignedUrlRequest body) {
     var bearerToken = bearerTokenFactory.from(request);
     var ip = request.getHeader("X-Forwarded-For");
-    var googleProject = request.getHeader("x-user-project");
+    var googleProject = body.getGoogleProject();
     var userAgent = request.getHeader("User-Agent");
     var signedUrl =
         signedUrlService.getSignedUrl(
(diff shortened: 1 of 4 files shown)


rjohanek (commented):
Hey Kevin, one idea that could make the logs easier to analyze when we get to that point - what if we only log if the x-project-header is non-null?

service/src/main/java/bio/terra/drshub/controllers/GcsApiController.java
snf2ye: Is the x-user-project the same as the body.getGoogleProject()? It looks like the user can specify a google project as a request parameter. 
I'm not really sure how we expect for the google project to be sent. Maybe we should log both?

service/src/test/java/bio/terra/drshub/controllers/DrsHubApiControllerTest.java
```diff
@@ -605,6 +610,35 @@ void testHandleEncodedSlashInDrsObjectId() throws Exception {
     postDrsHubRequestRaw(TEST_ACCESS_TOKEN, requestBody).andExpect(status().isOk());
   }
 
+  @Test
+  void testResolveDrsLogsRequest(CapturedOutput output) throws Exception {
```
snf2ye: I'm not sure tests are necessary for logging: There isn't really any logic we're testing and we'll still really need to check in prod that the logs are showing up as expected.

rjohanek:
> Hey Kevin, one idea that could make the logs easier to analyze when we get to that point - what if we only log if the x-project-header is non-null?

I just saw Florian's PR where he mentioned that you decided not to take this approach because you wanted to establish a baseline of all usages. That sounds good to me!

service/src/main/java/bio/terra/drshub/controllers/GcsApiController.java
kevinmarete: Yes, it is the same. It was added in this [PR](https://github.com/DataBiosphere/terra-drs-hub/pull/64)

snf2ye (approved):

rjohanek (approved):

PR #229, opened 2026-03-09
[DT-3011] Add retry logic for `x-user-project` header when required by TDR

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3011

### Summary

This PR adds logic in DrsHub to send the `x-user-project` header to TDR only when it is required.

DrsHub first requests without the header. If TDR returns a 400 Bad Request indicating that a header is required, DrsHub retries the request with the `x-user-project` header included.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -1,5 +1,6 @@
 package bio.terra.drshub.services;
 
+import static io.github.ga4gh.drs.model.Authorizations.SupportedTypesEnum.BEARERAUTH;
 import static org.apache.commons.lang3.ObjectUtils.isEmpty;
 
 import bio.terra.common.exception.BadRequestException;
@@ -12,6 +13,7 @@ import bio.terra.drshub.logging.AuditLogEvent;
 import bio.terra.drshub.logging.AuditLogEventType;
 import bio.terra.drshub.logging.AuditLogger;
 import bio.terra.drshub.models.AnnotatedResourceMetadata;
+import bio.terra.drshub.models.DrsApi;
 import bio.terra.drshub.models.DrsAuthEnum;
 import bio.terra.drshub.models.DrsHubAuthorization;
 import bio.terra.drshub.models.DrsMetadata;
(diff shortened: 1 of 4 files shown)


service/src/test/java/bio/terra/drshub/services/DrsResolutionServiceTest.java
```diff
@@ -342,4 +344,183 @@ void testDrsResolutionWithoutOptionalHeaders() throws Exception {
...
+                new AuditLogEvent.Builder(),
+                ip,
+                googleProject,
+                TRANSACTION_ID));
+
+    verify(drsApi, never()).setHeader(eq("x-user-project"), any());
+  }
+
+  @Test
+  void testNonTdrProviderAlwaysForwards() throws Exception {
```
rjohanek, line 488: this looks good. Just to clarify, there is also the case that the non TDR provider fails, and there is no retry logic associated with that in any case right?

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -327,6 +316,58 @@ AccessURL fetchDrsObjectAccessUrl(
...
+              uriComponents.toUriString(),
+              e.getMessage());
+          yield null;
+        }
+      }
+    };
+  }
+
+  private static boolean isRequireUserProjectError(HttpClientErrorException.BadRequest e) {
+    return e.getResponseBodyAsString().contains("Snapshot requires an x-user-project header");
```
rjohanek, line 361: It may be helpful to leave a comment in TDR where this error message is generated as a reminder that it is being relied on by DRS Hub so as to make this logic a little less brittle.

rjohanek (approved):
Overall, this looks good! Although we waiting to merge until we gather the logs, right? One minor suggestion about relying on the specific error message text from TDR. Other than that, did we do a full manual test to ensure this works as expected?

davidangb (approved):
ty!

service/src/test/java/bio/terra/drshub/services/DrsResolutionServiceTest.java
kevinmarete: Yes

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
fboulnois: I'll do that.

fboulnois (approved):
👍

PR #241, opened 2026-04-30
[CTM-473] Include bearer token with passport auth when googleProject is set

snf2ye:
https://broadworkbench.atlassian.net/browse/CTM-473

## Summary
When the DRS resolve endpoint uses passport auth and a `googleProject` parameter is provided, also send the bearer token alongside the passports. This enables signing the access URL with the userProject set with the correct access permissions.

## Implementation
- Searches the authorization list for a `BEARERAUTH` entry (from the fallback auth configuration) to get the properly configured token (either `provider_access_token` or `current_request`)
- Only includes the bearer token when `googleProject` is not null
- Added comprehensive logging to track when bearer tokens are included and warn if expected but not found

## Testing
- Added parameterized unit tests covering:
  - Bearer token is set when googleProject is provided
  - Bearer token is not set when googleProject is null  
  - No error when BEARERAUTH not in authorization list

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -370,12 +370,10 @@ public class DrsResolutionService {
             // Find BEARERAUTH authorization in the list to get the properly configured token
             Optional<String> bearerTokenOpt =
                 drsHubAuthorizations.stream()
-                    .filter(
-                        a -> a.drsAuthType() == Authorizations.SupportedTypesEnum.BEARERAUTH)
+                    .filter(a -> a.drsAuthType() == Authorizations.SupportedTypesEnum.BEARERAUTH)
                     .findFirst()
                     .flatMap(a -> a.getAuthForAccessMethodType().apply(accessMethodType))
-                    .flatMap(
-                        list -> list.isEmpty() ? Optional.empty() : Optional.of(list.get(0)));
+                    .flatMap(list -> list.isEmpty() ? Optional.empty() : Optional.of(list.get(0)));
             if (bearerTokenOpt.isPresent()) {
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -344,6 +360,31 @@ private static AccessURL callAccessUrl(
...
+          // signing the access url with the userProject set with the right access
+          if (googleProject != null) {
+            log.info(
+                "Google project {} specified for passport auth request to {}. Attempting to include bearer token.",
+                googleProject,
+                uriComponents.toUriString());
+            // Find BEARERAUTH authorization in the list to get the properly configured token
+            Optional<String> bearerTokenOpt =
+                drsHubAuthorizations.stream()
+                    .filter(a -> a.drsAuthType() == Authorizations.SupportedTypesEnum.BEARERAUTH)
```
davidangb, line 373: will `drsHubAuthorizations` actually contain a bearer token? I'm reading all the calling code, and I admit I don't understand all of it but it looks like it calls OPTIONS on the DRS provider - will TDR be responding and saying it supports bearer auth?

snf2ye: I think so because it tries the bearer token if the passport fails. 
And I tried using the OPTIONS endpoint on one of the prod DRS Ids and we get both:
```
"supported_types": [
    "PassportAuth",
    "BearerAuth"
  ]
```

davidangb: excellent, thank you!

davidangb (approved):

calypsomatic (approved):

PR #244, opened 2026-05-06
[CTM-472] use TDR client instead of ga4gh client for passport+requester-pays

davidangb:
https://broadworkbench.atlassian.net/browse/CTM-472
Use the TDR client instead of the ga4gh drs-spec client to invoke TDR when a request comes in with both a passport and a bill-to project for requester pays.

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -45,6 +45,7 @@ dependencies {
 	}
 	implementation 'org.broadinstitute.dsde.workbench:sam-client_2.13:v0.0.332'
 	implementation 'bio.terra:externalcreds-client-resttemplate:1.83.0-SNAPSHOT'
+	implementation 'bio.terra:datarepo-client:2.382.0-SNAPSHOT'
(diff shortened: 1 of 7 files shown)


service/build.gradle
```diff
@@ -45,6 +45,7 @@ dependencies {
 	}
 	implementation 'org.broadinstitute.dsde.workbench:sam-client_2.13:v0.0.332'
 	implementation 'bio.terra:externalcreds-client-resttemplate:1.83.0-SNAPSHOT'
+	implementation 'bio.terra:datarepo-client:2.382.0-SNAPSHOT'
```
snf2ye, line 48: Do we have a strategy for keeping this up to date? This gets bumped every time we merge to TDR. 
We probably should make sure dependabot is configured in way that updates this regularly but not on every version bump.

calypsomatic (approved):

snf2ye (approved):

PR #246, opened 2026-05-15
Revert "Revert "[CTM-472] use TDR client instead of ga4gh client for passport…"

davidangb:
Re-implements #244 (by reverting #245) to use Terra's TDR client instead of the GA4GH client when invoking PostAccessURL in TDR.

Adds logic on top of #244 to retrieve a pet token from Sam for the current user when making that PostAccessURL call.


davidangb:
Closing in favor of the `address-invalid-token` branch; the github action for drshub that deploys new images has problems with branches containing slashes.

PR #247, opened 2026-05-18
CTM-473: use TDR client library to call TDR's PostAccessURL

davidangb:
https://broadworkbench.atlassian.net/browse/CTM-473

https://github.com/DataBiosphere/terra-drs-hub/pull/247 and https://github.com/DataBiosphere/jade-data-repo/pull/2105 work together to support the (DRS resolution + passport + user-specified project for requester-pays) use case.

At a high level, these two PRs:
* In DRSHub, switch from using the GA4GH-standard DRS client library to the Terra-specific TDR client library for the TDR endpoint request in this specific use case. This allows passing the user's `Authorization: Bearer` token in the request.
* In TDR, add business logic to extract the bearer token from the request. We need this business logic because the TDR `/ga4gh/...` endpoint in question is missing some request headers that are only set by the oidc-proxy for requests under the `/api/...` prefix.

---

In this PR:
* Re-implement https://github.com/DataBiosphere/terra-drs-hub/pull/244 (by reverting https://github.com/DataBiosphere/terra-drs-hub/pull/245) to use Terra's TDR client instead of the GA4GH client when invoking PostAccessURL in TDR.

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/SamApiFactory.java
+++ b/service/src/main/java/bio/terra/drshub/services/SamApiFactory.java
@@ -3,6 +3,7 @@ package bio.terra.drshub.services;
 import bio.terra.common.iam.BearerToken;
 import bio.terra.drshub.config.DrsHubConfig;
 import bio.terra.sam.api.SamApi;
+import org.broadinstitute.dsde.workbench.client.sam.api.GoogleApi;
 import org.springframework.stereotype.Service;
 
 @Service
@@ -14,5 +15,12 @@ public record SamApiFactory(DrsHubConfig drsHubConfig) {
     samApi.getApiClient().setAccessToken(bearerToken.getToken());
 
     return samApi;
+  }
+
+  public GoogleApi getGoogleApi(String accessToken) {
+    var apiClient = new org.broadinstitute.dsde.workbench.client.sam.ApiClient();
+    apiClient.setBasePath(drsHubConfig.getSamUrl());
+    apiClient.setAccessToken(accessToken);
+    return new GoogleApi(apiClient);
   }
 }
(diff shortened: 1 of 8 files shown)


service/src/test/java/bio/terra/drshub/services/DrsResolutionServiceTest.java
```diff
@@ -532,17 +539,72 @@ private DrsProvider createTdrProviderWithRetryMode() {
...
+    verify(drsApi, never()).postAccessURL(any(), any(), any());
+    verify(drsApi, never()).setBearerToken(any());
   }
 
-  @ParameterizedTest
-  @MethodSource
-  void passportAuthWithGoogleProject(String googleProject, boolean shouldSetBearerToken)
-      throws Exception {
+  @Test
+  void passportAuthWithGoogleProject_usesTdrFactory() throws Exception {
```
calypsomatic, line 579: I think I added this test (or claude did), but it seems pretty redundant with the previous one

calypsomatic (approved):

service/src/main/java/bio/terra/drshub/services/SamApiFactory.java
```diff
@@ -15,4 +16,11 @@ public SamApi getApi(BearerToken bearerToken) {
 
     return samApi;
   }
+
+  public GoogleApi getGoogleApi(String accessToken) {
```
marctalbott: Is this used anywhere?

marctalbott (approved):

service/src/main/java/bio/terra/drshub/services/SamApiFactory.java
davidangb: thank you; removed!

PR #253, opened 2026-06-15
[CTM-537] add lat.md

calypsomatic:
https://broadworkbench.atlassian.net/browse/CTM-537

Adds claude-generated documentation based on recent issues as a helper for future developers and starter for further documentation.  Reviewers should check carefully that claude is not lying.

### DIFF
--- a/lat.md/drs-resolution.md
+++ b/lat.md/drs-resolution.md
@@ -20,7 +20,7 @@ Terra UI / Cromwell / WDS
 
 **Key distinction from normal auth:**
 - Normal: `GET /objects/{id}` with `Authorization: Bearer <terra-token>`
-- Passport: `POST /objects/{id}` with body `{"passports": ["<passport_jwt>"]}`
+- Passport: `POST /objects/{id}` with body `{"passports": ["<passport_jwt>"]}` — for metadata fetches only; for signed access URL requests with requester pays, both the passport and `Authorization: Bearer <terra-token>` are sent (see the requester-pays section below)
 
 ### Where it's configured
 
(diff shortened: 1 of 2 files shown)


lat.md/drs-resolution.md
```diff
@@ -0,0 +1,66 @@
...
+Terra UI / Cromwell / WDS
+    → DRSHub  POST /api/v4/drs/resolve
+         → ECM  GET /api/oidc/v1/{provider}/passport   (fetch stored RAS passport JWT)
+         → TDR  POST /ga4gh/drs/v1/objects/{id}        (fetch object metadata with passport)
+         → TDR  POST /ga4gh/drs/v1/objects/{id}/access/{accessId}  (get signed access URL)
+```
+
+**Key distinction from normal auth:**
+- Normal: `GET /objects/{id}` with `Authorization: Bearer <terra-token>`
+- Passport: `POST /objects/{id}` with body `{"passports": ["<passport_jwt>"]}`
```
davidangb: The critical change in https://github.com/DataBiosphere/terra-drs-hub/pull/247 was to send _both_ the passport and the `Authorization: Bearer <terra-token>` when calling the POST access url. So this line in the doc is misleading. This is explained at line 45 below, but here it could use a footnote or something

aednichols (approved):
Thank you!

snf2ye (approved):
+1 to David's comment, but otherwise LGTM

lat.md/drs-resolution.md
calypsomatic: Note added

PR #255, opened 2026-06-25
[CTM-472] Fix BDC DRS resolve returning accessUrl: null when userProject is present

snf2ye:
## Description

Fixes a bug where DRSHub returns `accessUrl: null` for BDC/Gen3 DRS URIs when the resolve request includes a `userProject`.

The `PASSPORTAUTH` branch in `callAccessUrl` was routing any passport-auth provider to the TDR client whenever `googleProject != null`. BDC doesn't use `x-user-project` and must not be routed to TDR — doing so caused TDR to return 400 for the unknown DRS ID, the Fence fallback to 401, and DRSHub to silently return `accessUrl: null`.

**Fix:** add a `supportsUserProject` flag to `ProviderAccessMethodConfig` (default `false`), set it `true` for TDR's `gs` method, and gate the TDR-client branch on it. Non-TDR providers now always use the standard passport POST path regardless of `googleProject`.

## Changes

- `ProviderAccessMethodConfigInterface` — new `supportsUserProject` flag (default `false`)
- `application.yml` — `supportsUserProject: true` on TDR's `gs` access method
- `DrsResolutionService` — gate TDR-client PASSPORTAUTH branch on `supportsUserProject`
- `DrsResolutionServiceTest` — fix 4 TDR tests to use TDR provider config; add BDC regression test

## Test plan

- [x] Unit tests pass (`DrsResolutionServiceTest`)
- [x] Confirmed that manual POST `/api/v4/drs/resolve` call w/ TDR URL still works
- [ ] Manual: POST `/api/v4/drs/resolve` for a BDC DRS URI with `userProject` returns a valid signed URL on dev tier - ⚠️ Note: I have not tested because I do not have a passport with access to a BDC consent code.

## References

CTM-472 / CTM-537

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
+++ b/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
@@ -296,6 +296,8 @@ public class DrsResolutionService {
     var accessMethodConfig = drsProvider.getAccessMethodByType(accessMethodType);
     boolean retryMode =
         accessMethodConfig != null && accessMethodConfig.requiresUserProjectOnRetry();
+    boolean supportsUserProject =
+        accessMethodConfig != null && accessMethodConfig.supportsUserProject();
 
     // Set x-user-project immediately for providers that always want it
     if (!retryMode && googleProject != null) {
@@ -316,7 +318,7 @@ public class DrsResolutionService {
                 uriComponents,
                 googleProject,
                 bearerToken,
-                retryMode);
+                supportsUserProject);
       } catch (HttpClientErrorException.BadRequest e) {
         if (retryMode && googleProject != null && isRequireUserProjectError(e)) {
           // Retry with a fresh client that includes x-user-project
@@ -333,7 +335,7 @@ public class DrsResolutionService {
                   uriComponents,
                   googleProject,
                   bearerToken,
-                  retryMode);
+                  supportsUserProject);
         } else {
           throw e;
         }
@@ -356,7 +358,7 @@ public class DrsResolutionService {
       UriComponents uriComponents,
       String googleProject,
       BearerToken bearerToken,
-      boolean requiresUserProjectOnRetry) {
+      boolean supportsUserProject) {
     Optional<List<String>> auth =
         authorization.getAuthForAccessMethodType().apply(accessMethodType);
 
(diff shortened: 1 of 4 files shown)


snf2ye:
I'm inclined to ignore the codecov results - as far I can tell, it complaining about code that I did not add in this PR.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
```diff
@@ -375,9 +380,14 @@ private AccessURL callAccessUrl(
...
         try {
-          // If googleProject is set, also send bearer token alongside passports to enable
-          // signing the access url with the userProject set with the right access
-          if (googleProject != null && bearerToken != null && bearerToken.getToken() != null) {
+          // If googleProject is set and this is a TDR provider, also send bearer token alongside
+          // passports to enable signing the access url with the userProject set with the right
+          // access. Non-TDR providers (e.g. BDC) do not use x-user-project and must not be routed
+          // to TDR.
+          if (supportsUserProject
+              && googleProject != null
```
aednichols, line 388: The semantics are a bit strange to me that we have both "can use a project" and "what is the project".

I wonder if it would be clearer to pass the repo identity and customize logic based on that?

e.g. 

```
if (repo = "terraDataRepo"
    && googleProject != null
    && [...]
```

aednichols (approved):
The reasoning is sound to me, left one optional comment about semantics, did not test.

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
snf2ye: Yeah, it's not great, but I do this it is reasonable line of questions - Can we use the project? If so, what is the provided project? 
And, unfortunately, checking if TDR is not sufficient since the https (aka Azure) option does not support the user project parameter.

davidangb (approved):

service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java
aednichols: Makes sense. Thanks for the explanation!

PR #258, opened 2026-07-28
CTM-613: Disambiguate cloud provider for TDR https-typed access methods

snf2ye:
# NOTE
I decided against this approach because I worried about unintended consequences. It's odd to me to use the "GS" access method when it really should be the "HTTPS" access method just to match the other GCP properties. It seems like this could get us in a wonky state.
____

## Summary
- Fixes [CTM-613](https://broadworkbench.atlassian.net/browse/CTM-613): resolving a TDR requester-pays DRS object via passport auth returned HTTP 500, because TDR's GCP passport signed-URL access method is typed `https` (same as Azure's), so the `supportsUserProject` gate added in #255 — keyed on `type` — never matched it, DRSHub sent the passport POST without a bearer token, TDR 401'd, and the bearer-GET fallback 500'd.
- Adds `DrsProviderInterface.getAccessMethodByTypeAndAccessId`, which resolves the provider's access-method config using the `az-` access-id prefix instead of `type` alone: for an `https`-typed access method, anything *not* `az-`-prefixed is treated as GCP and resolved against the provider's `gs` config. Non-`https` types (`gs`, `s3`, ...) are unaffected — the check only kicks in for the ambiguous `https` case, so it can't misroute other providers' `s3`/`gs` access methods.
- This is the DRSHub-side, access-id-prefix interim fix discussed on the ticket; it's self-contained and doesn't depend on any TDR-side change (e.g. adopting a `cloud` field).

## Why this approach
DRSHub already resolves the similiar ambiguity the same way in `AccessMethodUtils.getAccessMethodForCloud` — when a caller requests the Azure cloud platform, it matches on the `az-` access-id prefix rather than `type`, because a signed URL is `https` for every cloud and can't be told apart by `type` alone:
```java
if (cloudPlatform.equals(CloudPlatformEnum.AZURE)) {
  // Note: Only the Terra Data Repo drs provider prefixes Azure access ids with "az"
  filter = m -> m.getAccessId() != null && m.getAccessId().startsWith("az");
}
```

[CTM-613]: https://broadworkbench.atlassian.net/browse/CTM-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/bio/terra/drshub/config/DrsProviderInterface.java
+++ b/service/src/main/java/bio/terra/drshub/config/DrsProviderInterface.java
@@ -57,13 +57,14 @@ public interface DrsProviderInterface {
   /**
    * Like {@link #getAccessMethodByType}, but corrects for TDR returning its GCP passport signed-URL
    * access method typed {@code https} (the same type it uses for Azure) rather than {@code gs}.
-   * Only TDR prefixes GCP access ids with {@code gcp-}, so an access id with that prefix is
-   * resolved against the provider's {@code gs} config instead, regardless of the type DRS reported.
-   * See CTM-613.
+   * Only TDR prefixes Azure access ids with {@code az-}; any other {@code https} access id is
+   * treated as GCP and resolved against the provider's {@code gs} config instead, regardless of the
+   * type DRS reported. See CTM-613.
    */
   default ProviderAccessMethodConfig getAccessMethodByTypeAndAccessId(
       AccessMethod.TypeEnum accessMethodType, String accessId) {
-    if (accessId != null && accessId.startsWith("gcp-")) {
+    boolean isAzure = accessId != null && accessId.startsWith("az-");
+    if (accessMethodType == AccessMethod.TypeEnum.HTTPS && !isAzure) {
       var gcpConfig = getAccessMethodByType(AccessMethod.TypeEnum.GS);
       if (gcpConfig != null) {
         return gcpConfig;
--- a/service/src/main/java/bio/terra/drshub/config/DrsProviderInterface.java
+++ b/service/src/main/java/bio/terra/drshub/config/DrsProviderInterface.java
@@ -64,8 +64,8 @@ public interface DrsProviderInterface {
    */
   default ProviderAccessMethodConfig getAccessMethodByTypeAndAccessId(
       AccessMethod.TypeEnum accessMethodType, String accessId) {
-    boolean isGcpPassport = accessId != null && accessId.startsWith("gcp-");
-    if (accessMethodType == AccessMethod.TypeEnum.HTTPS && isGcpPassport) {
+    boolean isGcp = accessId != null && accessId.startsWith("gcp-");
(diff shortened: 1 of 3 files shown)


snf2ye (commented):
Ran a code review over this diff. Two follow-ups worth tracking (not blocking this PR, since both are currently masked by config) and one small cleanup in the new code.

### 1. `AuthService`'s `BEARERAUTH` path is still accessId-unaware
[AuthService.java#L154](https://github.com/DataBiosphere/terra-drs-hub/blob/b614246f8f86683c00516e26037d0c2fcf5200b0/service/src/main/java/bio/terra/drshub/services/AuthService.java#L154) (also [L164](https://github.com/DataBiosphere/terra-drs-hub/blob/b614246f8f86683c00516e26037d0c2fcf5200b0/service/src/main/java/bio/terra/drshub/services/AuthService.java#L164))

This resolves auth via `drsProvider.getAccessMethodByType(accessType)`, not the new `getAccessMethodByTypeAndAccessId`. For a TDR GCP-passport object (typed `https`, accessId prefixed `gcp-`), if TDR's `Authorizations` OPTIONS response ever lists `BEARERAUTH` as a supported type, and `gs`/`https` `auth`/`fallbackAuth` configs are ever set differently (today they happen to match), this would resolve using the wrong (`https`) config — reintroducing the same class of bug CTM-613 fixes elsewhere.

### 2. `DrsProviderInterface.shouldFetchAccessUrl` is also type-only, not accessId-aware
[DrsProviderInterface.java#L82-L96](https://github.com/DataBiosphere/terra-drs-hub/blob/b614246f8f86683c00516e26037d0c2fcf5200b0/service/src/main/java/bio/terra/drshub/config/DrsProviderInterface.java#L82-L96)

It decides whether to fetch a signed URL by scanning *all* access-method configs for any `https`-typed one with `fetchAccessUrl: true`, rather than resolving the specific config for this object via `getAccessMethodByTypeAndAccessId`. In `application.yml`, TDR's `gs.fetchAccessUrl` and `https.fetchAccessUrl` both happen to be `true` today, so this is masked. If `https.fetchAccessUrl` were ever set to `false` (e.g. to stop fetching signed URLs for Azure), DRSHub would silently stop fetching access URLs for GCP-passport objects too, since they're also typed `https`.

### 3. Minor: redundant recompute in the new code
[DrsResolutionService.java#L346](https://github.com/DataBiosphere/terra-drs-hub/blob/b614246f8f86683c00516e26037d0c2fcf5200b0/service/src/main/java/bio/terra/drshub/services/DrsResolutionService.java#L346)

`getAccessMethodByTypeAndAccessId(accessMethodType, accessId)` is called again here with the same args already computed into `accessMethodConfig` a few lines up (L296-297) — could just use `accessMethodConfig.getAuth()`. Not a bug, just avoids a redundant list scan.

